PR #2, opened 2019-12-25
Project 13 App with React Native Angelica B

Angelicabrodin:
This week has been one of the most stressful week at work before Christmas, therefore I decided quite early to give myself some slack and finish this project doing my best, based on these circumstances. 

And I had a really hard time getting the npm installed, which also made me a bit stressed in the beginning, but eventually I got everything to work, except the IoS simulator, but managed to use tunnel in expo when scanning the QR-code and that worked out OK :)

I have more or less watched all the videos, and followed the code alongs provided, and decided to do the APP as Damien shows in the first video. I thought it was quite hard to wrap my head around, especially how to use the different components within other components, such as for instance putting a "div" inside another "div" which in this case would be a container/view and then a text component. I tried this a lot, since I really wanted to add some extra details to the movies. But could not make it work, so in the end decided to leave it, so maybe you could give me some tips on how to do that :)

Another thing that I struggled with, was the inspector tool, therefore I also really struggled with the styling, such as margins and positioning, because I really did not understand the "boxes". 

Also, I could not take away the "adult movies" haha, so a lot of interesting movie selections coming up in my feed :P what am I doing wrong? 

And finally, the ActivityHandler is so far up on the page it is barely visible, so maybe some tips on how to style that as well :)

I think overall this is a really fun and creative project, just for me, on the worst week of the year, and I know that I could work on it over Christmas, but I have decided to give myself a well deserved break, so here we go, I am handing it in :)

So thats all for me,
Over and out/
Angelica

### DIFF
--- a/App.js
+++ b/App.js
@@ -8,18 +8,21 @@ const TOP_MOVIE_ID = 656516
 const Container = styled.View`
   flex: 1;
   background-color: black;
-  align-items: center;
-  justify-content: center;
+  align-items:center;
+
 `
 const Button = styled.TouchableOpacity`
-background:green;
-position:absolute;
-bottom:40;
-padding:10px 20px;
-border-radius:20px;
+  background:gray;
+  position:absolute;
+  bottom:70;
+  margin:20px;
+  padding:10px 20px;
+  border-radius:30px;
+  width:50%;
 `
 const ButtonText = styled.Text`
   color:white;
+  font-weight:bold;
   font-size:18;
 `
 
@@ -36,6 +39,7 @@ const fetchRandomMovie = async () => {
 }
 
 const App = () => {
+
   const [movie, setMovie] = useState()
   const [loading, setLoading] = useState(true)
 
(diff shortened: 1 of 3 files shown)


App.js
```diff
@@ -1,26 +1,70 @@
...
+const TOP_MOVIE_ID = 656516
 
 const Container = styled.View`
   flex: 1;
-  background-color: papayawhip;
-  justify-content: center;
-  align-items: center;
-`
+  background-color: black;
+  align-items:center;
```
JennieDalgren: To make sure that the `ActivityIndicator` shows, add some padding to the top of this container
` padding-top: 60px;` for example.

App.js
```diff
@@ -1,26 +1,70 @@
...
+    fetchMovieData()
+  }, [])
 
-export default App
+  return (
+    <Container>
+      {loading
+        ? <ActivityIndicator size="large" color="white" />
+        : <Movie movie={movie} />}
+      <Button onPress={fetchMovieData}>
```
JennieDalgren: Great job with this logic!

App.js
```diff
@@ -1,26 +1,70 @@
...
+  const [movie, setMovie] = useState()
+  const [loading, setLoading] = useState(true)
+
+  const fetchMovieData = () => {
+    setLoading(true)
+    fetchRandomMovie()
+      .then((movieData) => {
+        setMovie(movieData)
+        setLoading(false)
+      })
```
JennieDalgren: Nice! Also great with the loading changing true/false 💪🏼

PR #3, opened 2019-12-27
Project React-native-app - Susanna Sjöberg

SanSjo:
This project was hard in the beginning, but also very du  and interesting. I'm very happy that I have an app that works now with the requirements. 
I wont be able to join the meetup #3 so unfortunately i cant demo my app, but maybe I can do it  later or in another way.


App.js
```diff
@@ -1,26 +1,116 @@
...
+    const photo = photos.find(p => p.id === photoId);
+    return photo.src.original;
+  };
+
+  const onShare = async photoId => {
+    const result = await Share.share({
+      message: 'Hey! Look at this beautiful photo from Pexels',
+      url: getImageToShare(photoId),
+      title: 'Wow, did you see that?'
+    });
```
JennieDalgren: Nice job with the Share!!

components/Header.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { StyleSheet, View, Text } from 'react-native';
+
```
JennieDalgren: I see that you in this component used **css in js** instead of **styled components.** I guess that's because you wanted to try that out as well. That is totally fine for this project, but going forward try to stick to one way of styling your components.

PR #6, opened 2020-01-05
React Native App - Linda Isell

liissee:
Link to snack: https://snack.expo.io/@liissee/testing-todo-app

Reflections: 
Everything felt so easy before I started with the app and I think I started at least 5 different projects before I decided to go with a simple to-do app.. 
I felt that it was hard to follow and use the React Native documentation. I tried to do a step-counter but I failed to translate the "class" example to "hooks". Would love to get more examples when translating class to hooks.

Regarding styled components: sometimes you need to import "Text" and "View" and sometimes it said that it was already declared. For example I had some trouble with "Scrollview", I tried to style it with styled components but then I needed to style it within the component to make it work..? 
Sorry for the bad explanation, hope you understand what I mean :)


App.js
```diff
@@ -1,26 +1,150 @@
...
   return (
-    <Container>
-      <Title>This is your cool app!</Title>
-      <Title>Go to App.js and start coding</Title>
-      <Title>💅💅💅</Title>
-    </Container>
+    <SafeAreaView>
+      <ImageBackground
+        source={require("./assets/planner.png")}
+        style={{ width: "100%", height: 120 }}>
```
JennieDalgren: try to keep the styling away from here if possible.
style in styled components or with the styleSheet

App.js
```diff
@@ -1,26 +1,150 @@
...
+      <ScrollView style={{ marginHorizontal: 20, width: "100%" }}>
+        {todos.map(item => (
+          <TodoList
+            text={item.text}
+            key={item.key}
+            checked={item.checked}
+            setChecked={() => checkTodo(item.key)}
+            deleteTodo={() => deleteTodo(item.key)}
+          />
+        ))}
```
JennieDalgren: ⭐️

PR #10, opened 2020-01-06
React-native-project-Disa

Disa-san:
Snack link: https://snack.expo.io/@disa-san/todo-list---disa

I relied heavily on the knowledge of the internet for this task ;-). As much as I liked the idea and freedom of this project I found it very difficult. I started many different apps before finally settling with this one. I wanted to build a thesaurus app, that could be used to get alternative words when writing, but could not wrap my head around how to use the API that I found. 

I have not yet got the grip on React or React Native and I feel that it is a slow learning curve for me. I find solutions to my problems, but I don't know why it works and sometimes get stuck on the most basic things, or so it seems. I still don't understand how to properly use props (no pun intended), although I have come as far as to understand why it should be used.

### DIFF
--- a/App.js
+++ b/App.js
@@ -1,9 +1,12 @@
 import React, { useState } from 'react'
-import { StyleSheet, Text, View, TouchableOpacity, TextInput, ScrollView } from 'react-native'
+import { TouchableOpacity, ScrollView } from 'react-native'
+import styled from "styled-components"
 
 import Icon from 'react-native-vector-icons/Feather'
 
 import TodoList from './Components/TodoList'
+
+
 
 export default function App() {
   const [value, setValue] = useState('')
@@ -34,11 +37,11 @@ export default function App() {
   }
 
   return (
-    <View style={styles.container}>
-      <Text style={styles.header}>Todo List</Text>
-      <View style={styles.textInputContainer}>
+    <Container>
+      <Title>Todo List</Title>
+
+      <TextInputContainer>
         <TextInput
-          style={styles.textInput}
           multiline={true}
           placeholder="Feed me Todos"
           placeholderTextColor="grey"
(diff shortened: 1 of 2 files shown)


App.js
```diff
@@ -1,26 +1,111 @@
...
+
+export default function App() {
+  const [value, setValue] = useState('')
+  const [todos, setTodos] = useState([])
+
+  addTodo = () => {
+    if (value.length > 0) {
+      setTodos([...todos, { text: value, key: Date.now(), checked: false }])
+      setValue('')
+    }
```
JennieDalgren: nice to check if the input is not empty before saving the todos

App.js
```diff
@@ -1,26 +1,111 @@
...
+  }
+
+  checkTodo = id => {
+    setTodos(
+      todos.map(todo => {
+        if (todo.key === id) todo.checked = !todo.checked
+        return todo
+      })
```
JennieDalgren: nice!

App.js
```diff
@@ -1,26 +1,111 @@
...
+    )
+  }
+
+  deleteTodo = id => {
+    setTodos(
+      todos.filter(todo => {
+        if (todo.key !== id) return true
+      })
+    )
```
JennieDalgren: nice, this could also be shortened to
`setTodos(todos.filter(todo => todo.key !== id)`

App.js
```diff
@@ -1,26 +1,111 @@
...
+
+import TodoList from './Components/TodoList'
+
+
+
+export default function App() {
+  const [value, setValue] = useState('')
+  const [todos, setTodos] = useState([])
+
+  addTodo = () => {
```
JennieDalgren: make sure to add the keyword `const` before creating the functions.

`const addTodo = () => { ... }`

Components/TodoList.js
```diff
@@ -0,0 +1,59 @@
+import React from 'react'
+import { StyleSheet, View, Text } from 'react-native'
+import Icon from 'react-native-vector-icons/Feather'
+import styled from 'styled-components'
+
+export default function TodoList(props) {
+  return (
+    <ListContainer>
+      <Icon
+        name={props.checked ? 'check' : 'square'}
```
JennieDalgren: nice little ternary here, to show the correct icon.
This is a good use of props, for example.

PR #12, opened 2020-01-07
React Native App - Jenny Jacobsson

jennyjacobsson:
It was fun building my very first app, but also a bit tricky. I had some styling that I used in multiple components, so I added a 'styledCollection' file to store them all in. I don't know if that's the right way, but it seemed better than writing the same code multiple times. Is there a better way to do it? Also I’ve only been able to work with iOS so I have no idea if the app works on Android.

See you soon!


App.js
```diff
@@ -1,26 +1,141 @@
...
+
+  const { image_url } = giphy;
+
+  if (loading) {
+    return (
+      <Container>
+        <ActivityIndicator size="large" color="#ffffff" />
+        <GiphyLogo />
+      </Container>
+    );
```
JennieDalgren: nice little loading state!

components/GifView.js
```diff
@@ -0,0 +1,69 @@
...
+        />
+      </ImageWrap>
+      <BottomButtons>
+        <ShareButton image_url={image_url} />
+        <Button
+          onPress={() => {
+            Clipboard.setString(image_url);
+            setCopied(true);
+            setTimeout(() => {
+              setCopied(false);
```
JennieDalgren: nice to use the clipboard functionality

components/GifView.js
```diff
@@ -0,0 +1,69 @@
...
+            }, 3000);
+          }}
+          active={copied}
+        >
+          <ButtonText>{copied ? "Copied" : "Copy"}</ButtonText>
```
JennieDalgren: nice little ternary here ⭐️

PR #13, opened 2020-01-07
React Native app - Emmie Nilsson

eilsson:
I made a simple app that displays a random Pokémon. I'm not happy about how this project turned out. I struggled quite a bit with making Expo work and styling. 
I managed to split split my components into smaller ones. I wanted to add information about what type the pokémon was and what types of pokémon its effective and not that effective against, but I had a lot of trouble with props in this project so I skipped that. Might try to add that later on.


assets/components/Title.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
```
JennieDalgren: nice way of creating a common component that you'll reuse many times! ⭐️

assets/components/PokemonCard.js
```diff
@@ -0,0 +1,35 @@
+import React from "react";
```
JennieDalgren: neat and clean! perfect

assets/components/LoadingPokemonCard.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import { ActivityIndicator, View } from "react-native";
+import styled from "styled-components";
+
+const LoadingPokemonCard = () => {
```
JennieDalgren: Maybe to make this component even more general it could be named only Loading or LoadingSpinner?
Then you could reuse it in any project you want without having to change anything to it.

App.js
```diff
@@ -1,26 +1,71 @@
...
+import Title from "./assets/components/Title";
+import CustomButton from "./assets/components/CustomButton";
+import LoadingPokemonCard from "./assets/components/LoadingPokemonCard";
+import PokemonCard from "./assets/components/PokemonCard";
+
+const randomizePokemon = (min, max) => {
+  min = Math.ceil(min);
+  max = Math.floor(max);
+  return Math.floor(Math.random() * (max - min) + min); // max value is exclusive and min value inclusive, so the value is no lower than min but less than and not equal to max
+};
```
JennieDalgren: nice randomizer.
This could even be extracted into its own file, and made general to be used in other projects.
Then just import it here and use it as you do on line 18

PR #15, opened 2020-01-07
React Native App - Agnes

Agnes94:
I kept it simple and created a note taking app. I wasn't very inspired so it's pretty basic, but I had fun while doing it. Will definitely need a lot more time to wrap my head around react native in general. I also found it very hard to focus during the holiday season.


components/addTodo.js
```diff
@@ -0,0 +1,57 @@
...
+import {
+  StyleSheet,
+  Text,
+  View,
+  TextInput,
+  TouchableOpacity
+} from "react-native";
+
+export default function AddTodo({ submitHandler }) {
+  const [text, setText] = useState("");
```
JennieDalgren: nice work with the props

App.js
```diff
@@ -1,26 +1,65 @@
...
+export default function App() {
+  const [todo, setTodo] = useState([
+    { text: "Buy ginger", key: "1" },
+    { text: "Drink tea", key: "2" },
+    { text: "Go to the park", key: "3" }
+  ]);
+
+  const pressHandler = key => {
+    setTodo(prevTodo => {
+      return prevTodo.filter(todo => todo.key != key);
```
JennieDalgren: use !== instead, it's more strict

components/todoItem.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import { StyleSheet, Text, TouchableOpacity } from "react-native";
+
+export default function TodoItem({ item, pressHandler }) {
```
JennieDalgren: ⭐️

PR #23, opened 2020-01-09
WildBerry by Nasim Mahzoun

Nasimmhn:
The app's name "WildBerry" refers to "Wild Strawberries", and the Swedish word "Smultronställe".
The app is just a demo. You can save new places but when you restart the app the added places are gone - since the new places are saved in an array and not in a database.
When you add a new place you type the street and city, and then an API will return the latitude and longitude in order to display the new place with a marker on the map.
There are also some other things to fix, e.g. assigning icons to added places.


components/FavMarkers.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { Marker } from 'react-native-maps'
```
JennieDalgren: Nice that you added this as its own component. Also that you found out how to use it!

components/FindLocation.js
```diff
@@ -0,0 +1,12 @@
+import { LOCATION_API } from 'react-native-dotenv'
+
+
+const FindLocation = async (search) => {
+
+  // const uri = `https://maps.googleapis.com/maps/api/geocode/json?address=${search}&key=${GOOGLE_API}`
+  const uri = `https://eu1.locationiq.com/v1/search.php?key=${LOCATION_API}&q=${search}&format=json`
```
JennieDalgren: I can't get this LOCATION_API to work when I run this on my computer...

models/place.js
```diff
@@ -0,0 +1,12 @@
...
+    this.id = id;
+    this.title = title;
+    this.description = description;
+    this.position = position;
+    this.location = location
+    this.icon = icon
+  }
+}
+
+export default Place;
```
JennieDalgren: What happened here with using a class?

navigation/PageNavigator.js
```diff
@@ -0,0 +1,27 @@
+import { createAppContainer } from 'react-navigation';
+import { createStackNavigator, TransitionPresets } from 'react-navigation-stack';
```
JennieDalgren: Nice that you tried out the react navigation! 💪🏼

screens/AddPlaceScreen.js
```diff
@@ -0,0 +1,233 @@
...
+  const json = await res.json()
+  return json
+}
+
+export const AddPlaceScreen = (navData) => {
+
+  const [title, setTitle] = useState("")
+  const [description, setDescription] = useState("")
+  const [street, setStreet] = useState("")
+  const [city, setCity] = useState("Stockholm")
```
JennieDalgren: ⭐️

screens/AddPlaceScreen.js
```diff
@@ -0,0 +1,233 @@
...
+
+
+        <View style={styles.iconsContainer}>
+          <IconButton>
+            <Ionicons
+              color={"#fff"}
+              style={styles.icon}
+              name="ios-home"
+            />
+          </IconButton>
```
JennieDalgren: You could make this into its own component and have props like name and color. Where you also have a default color to white so you don't have to specify everytime.
`<Icon name="ios-home" /> `and then just add this for each icon button

screens/MapScreen.js
```diff
@@ -0,0 +1,84 @@
+import React from 'react'
```
JennieDalgren: Nice with the mapview! 🎉

PR #25, opened 2020-01-12
Blue Sky News

ninamansster:
IMG_5754
IMG_5755
IMG_5756

Hi, here is my Blue Sky News App.
I have used Styled components and in one component I have stylesheets, so I have elaborated on both techs.
Had some issues with the Picker on iOS it works better on Android.
All the best/
Nina


App.js
```diff
@@ -1,26 +1,191 @@
...
+
+
+        <Image
+          source={require('./assets/bluesky.jpg')}
+          style={{ width: 350, height: 150 }} />
+        <View1>
+          <Title4>Welcome to Blue Sky News!</Title4>
+          <Title4>Updated from the world {moment().calendar()}</Title4>
+          <Picker
+            style={{ height: 50, width: 150, margin: 10, }}
```
JennieDalgren: try to not use inline styling like this.
Stick to styled components or StyleSheet.

App.js
```diff
@@ -1,26 +1,191 @@
...
+  font-weight: 700;
+  padding-top: 50px;
+`
+const Title1 = styled.Text`
+  font-size: 30;
+  color: black;
+  padding-bottom: 15;
+  font-weight:800;
+`
+const Title2 = styled.Text`
```
JennieDalgren: try to use more explanatory names on you components.

components/CustomButton.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
```
JennieDalgren: Nice that you created a button component in it's own file like this anbd then importing it where you need it ⭐️

PR #26, opened 2020-01-14
Project React Native App, Michel Elmoznino Laufer

MichelLaufer:
I think that this week's project was fun but I initially had some technical difficulties to get started with the project with Expo "crashing" several times when I tried to scan the QR-code with my phone. But once I got started and decided what type of app that I wanted to do (a movie quiz app), I thought it was fun.

Link to Expo snack: https://snack.expo.io/@michellaufer/2ea9f9


App.js
```diff
@@ -1,26 +1,326 @@
...
+  const rightAnswer = () => {
+    setScore(score+1)
+    setHide(true)
+    setRight(true)
+  }
+
+  const wrongAnswer = () => {
+    setHide(true)
+    setWrong(true)
+    return Vibration.vibrate(duration)
```
JennieDalgren: nice that you tried the Vibration functionality

App.js
```diff
@@ -1,26 +1,326 @@
...
+      {question === 'begin' && (
+      <View>
+        <Container>
+        <Title>Welcome to Michels movie quiz</Title>
+        <Title>🍿🍿🍿</Title>
+        <Title>❓❓❓</Title>
+        <Button>
+          <ButtonText onPress={() => setQuestion('one')}>Start the quiz</ButtonText>
+        </Button>
+        <LottieView source={require('./1879-movie-loading.json')} autoPlay loop style={{height: 100}}/>
```
JennieDalgren: Yay! You managed to add a Lottie animation. This is super useful knowledge if you're getting into more react native projects.

App.js
```diff
@@ -1,26 +1,326 @@
...
+        )}
+        {right && (
+          <TitleRight>Correct answer! Please go to the next question</TitleRight>
+        )}
+        {wrong && (
+          <TitleWrong>Wrong answer! Please go to the next question</TitleWrong>
+        )}
+        <Score>{score} points</Score>
+        </Container>
+      </View>
```
JennieDalgren: You could make this into a question component and import it with the props:
title, question, rightAnswerIndex or similar and then import a <Question /> here instead of duplicating the code.

App.js
```diff
@@ -1,26 +1,326 @@
...
+          </Button>
+          <Button>
+            <ButtonText onPress={wrongAnswer}>Johnny Depp</ButtonText>
+          </Button>
+          <Button>
+            <ButtonText onPress={wrongAnswer}>Tom Cruise</ButtonText>
+          </Button>
+          <Button>
+            <ButtonText onPress={wrongAnswer}>Daniel Craig</ButtonText>
+          </Button>
```
JennieDalgren: This could be extracted further into the "options component" and here you could use checkboxes or radio buttons then do a function to check if the selected one is the correct one. I don't know if you want me to explain about this further, or if this is enough? (based on your Slack comment)

PR #36, opened 2020-04-07
React Native App - Advice App - RAGNA

Ragtulf:
Would have liked to spend more time on animation but I'm content.


assets/Components/ShakeEventExpo.js
```diff
@@ -0,0 +1,28 @@
+import { Accelerometer } from 'expo-sensors';
```
JennieDalgren, line 1: Nice that you got this to work, its totally fine to use copy pasted code like this when you're trying things out and learning for the first time. If you have time and energy, try to refactor this into a functional component :)

assets/Components/TouchableButton.js
```diff
@@ -0,0 +1,16 @@
...
+    <View>
+      <TouchableHighlight
+        onPress={onPress}
+      >
+        <ButtonText>{text}</ButtonText>
+      </TouchableHighlight>
+    </View>
+
+  )
+}
```
JennieDalgren, line 16: Perfect way of building a reusable component. Although, try to stick to the code guide lines we're used to.
```
export const TouchableButton = () => (
  <View>...</View>
)
```

App.js
```diff
@@ -1,25 +1,45 @@
...
 
 const App = () => {
+  const [firstView, setFirstView] = useState(true)
+
+  const shakePhone = () => {
+    setFirstView(false)
+  }
+
+  useEffect(() => {
+    ShakeEventExpo.addListener(() => {
```
JennieDalgren, line 24: 😎

PR #37, opened 2020-04-08
Review Wanted: Week13 - React Native App: Bea

bealun:
(Made my app in expo first that's why all of the code suddenly appear in the first commit :))
Started out with the layout of the form to have one question with different answers to chose from per screen. If-statments to show the right question and then functions that will handle the answers. I've had a hard time structuring without losing some info on the way - which means that all of my code is in the App.js file, unfortunately. Overall happy with the result and the the fact that I completely understand the code I've written.

Expo Snack: https://snack.expo.io/@bealun/office-quiz

**Is there a way to make a React Native app (or apps in general) accessible?**


App.js
```diff
@@ -1,26 +1,364 @@
...
+const Score = styled(Title)`
+  padding-top: 10px;
+  font-size: 12px;
+`;
+
+// const sumAnswer = styled(Title)`
+//   border: 1px solid;
+//   font-size: 16px;
+//   padding-bottom: 2px;
+// `;
```
JennieDalgren, line 319: remove unused code if there isn't a valid point why to keep it.

components/Styled.js
```diff
@@ -0,0 +1,117 @@
+// import styled from 'styled-components/native'
```
JennieDalgren, line 1: How come you didn't use this file?
Looks like a good idea to extract all the styling into it's own file and then import the styled components where you want them?
just remember the export
`export const Container = styled.View`

JennieDalgren: if not, maybe delete this file instead.

App.js
```diff
@@ -1,26 +1,364 @@
...
+  const rightAnswer = () => {
+    setScore(score + 1);
+    setHide(true);
+    setRight(true);
+  };
+
+  const wrongAnswer = () => {
+    setHide(true);
+    setWrong(true);
+    return Vibration.vibrate(duration);
```
JennieDalgren, line 50: nice that you're trying out the vibrate functionality.

PR #38, opened 2020-04-08
rewie wanted

kajsabootin:
I think it was really hard to use style-component, but also really fun.

I am most proud of that I have done my first app. 

 Maybe have some more video tutorials about style-component.


App.js
```diff
@@ -1,26 +1,37 @@
...
-      <Title>Go to App.js and start coding</Title>
-      <Title>💅💅💅</Title>
+
+        <Header/> 
+        
+        <View>
+
+        <MagicBall/>
+
+        </View>
```
JennieDalgren, line 17: indentation!
Do you actually need the View around MagicBall?

MagicBall.js
```diff
@@ -0,0 +1,143 @@
...
+  {
+    answear: "It is certain.",
+  },
+  {
+    answear: "It is decidedly so.",
+  },
+  {
+    answear: "Without a doubt.", 
+  },
+  {
```
JennieDalgren, line 14: since you only have one key-value pair in each object maybe it's enough to have the answers just as strings in the array?
such as:
```
const quotes = [
  'It is certain',
  'Without a doubt',
  '...'
]
```
Also, stick to either `const quotes` or `const QUOTES` it'd not a component so dont use Quotes with capital Q.

MagicBall.js
```diff
@@ -0,0 +1,143 @@
...
+        <SmallCircel>
+          <Answear> {quote.answear} </Answear>
+        </SmallCircel>
+      </BigCircel>
+
+      <TouchableOpacity onPress={showQuote}>
+      <Text>Ask me</Text>
+      </TouchableOpacity>
+      
+    </View>
```
JennieDalgren, line 91: Indentation!

MagicBall.js
```diff
@@ -0,0 +1,143 @@
...
+  font-size: 25;
+  text-align: center;
+  font-weight: bold;
+`
+const TouchableOpacity = styled.TouchableOpacity`
+  background-color: #FCED84;
+  border-radius: 50;
+  padding-left: 15;
+  padding-right: 15;
+  margin-top: 100;
```
JennieDalgren, line 142: I think this causes some issues that you can't really see the button on some devices. I have an iPhone X and this is how it looks.
Try to not add a lot of extra margin or padding. Use flex styling instead, such as 
`justify-content: space-between` or something like that to "spread" the components out.

JennieDalgren: IMG_6710

JennieDalgren: oops, it became super big!

PR #39, opened 2020-04-08
Review wanted - Frida Eriksson

fridamaria:
My reflections on how this week's project turned out: it was fun to see how we can implement what we've learned in a different setting. I tried looking at tutorials and adjusted them to what I know since before, utilizing some of the logic we have worked with for previous projects. I decided to keep it simple though and give myself a bit of a break, so decided to not do any stretch goals this week. 

Things I'd like to have clarified or explained in more detail: I would like to know more about how to store the information in the app, but I guess it will become easier when we learn how to create our own API's.

I made a copy of my project and added it to snack in case anyone wants to have a look there: https://snack.expo.io/@fridamaria/todo-app


components/NewTodo.js
```diff
@@ -0,0 +1,62 @@
...
+        value={newTodo}
+        onChangeText={handleChange}
+        placeholder='Add task'
+        maxLength={25}
+      />
+
+      <AddButton
+        onPress={() => {
+          handleSubmit(newTodo)
+          Vibration.vibrate()
```
JennieDalgren, line 52: Nice with the vibration functionality

components/NewTodo.js
```diff
@@ -0,0 +1,62 @@
...
+          setNewTodo('')
+        }}
+        disabled={newTodo.length < 3 || newTodo.length > 25 ? true : false}
```
JennieDalgren, line 55: 😎 👍

components/TodoItem.js
```diff
@@ -0,0 +1,62 @@
...
+  ${prop => prop.completed ? 'color: #bbb; text-decoration: line-through; text-decoration-color: #bbb' : 'color: #fff'};
+`
+
+export const TodoItem = ({ text, item, handleBinPress }) => {
+  const [completed, setCompleted] = useState(false)
+
+  //Function used to toggle if a todo item is completed or not, used to render different styling
+  const toggleCompleted = () => {
+    { completed ? setCompleted(false) : setCompleted(true) }
+  }
```
JennieDalgren, line 46: you could also do:
`setCompleted(!completed)`
that will "invert" the value, so if it's false, will be set to true and vice versa.

App.js
```diff
@@ -1,26 +1,72 @@
...
+
   return (
-    <Container>
-      <Title>This is your cool app!</Title>
-      <Title>Go to App.js and start coding</Title>
-      <Title>💅💅💅</Title>
-    </Container>
+    <OuterContainer>
+
+      <StatusBar barStyle="light-content" />
```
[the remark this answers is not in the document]
fridamaria: If you mean the StatusBar barStyle it should change the content of the status bar so that the text etc. is light

components/TodoItem.js
fridamaria: Ah, nice!

fridamaria:
Thank you for your feedback Jenny ⭐

PR #40, opened 2020-04-08
Project React Native App Anna Takman

annatakman:
I went for a step counter app, mainly because I thought it would be fun to use one of the functionalities in the phone, in this case the Pedometer. It was a really fun project!


components/StepCounter.js
```diff
@@ -0,0 +1,58 @@
...
+
+const Text = styled.Text`
+  font-size: 18px;
+  margin: 28px;
+  font-weight: bold;
+  color: #182C61;
+`
+const Container = styled.View`
+justify-content: center;
+align-items: center;
```
JennieDalgren, line 19: indentation

components/Animation.js
```diff
@@ -0,0 +1,11 @@
+import React from "react"
+import LottieView from "lottie-react-native"
+
+
+const Animation = () => {
+  return (
+    <LottieView style={{ width: 350, height: 350, backgroundColor: "#F8EFBA" }} source={require('../assets/walking.json')} autoPlay loop />
+  )
+}
```
JennieDalgren, line 9: nice that you used a Lottie animation! Makes such a big difference having a nice animation like that! ⭐

components/StepCounter.js
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <Container>
+      <Title>Steps taken the last week: {weekStepCount}</Title>
+      <Title>Steps taken the last 24 hours: {pastStepCount}</Title>
+      <Text>Walk and watch this go up: {currentStepCount}</Text>
+    </Container>
+  )
+}
+
```
JennieDalgren, line 57: really nice work adding this Pedometer! 👯‍♀️

PR #46, opened 2020-04-10
Quote generator

elinaarenstrup:
My reflections on how this week's project turned out: It was fun and challenging.

Things I'd like to have clarified or explained in more detail:  How do I "deploy" a React native app?


App.js
```diff
@@ -1,26 +1,34 @@
-import React from 'react'
-import styled from 'styled-components/native'
+import React from "react";
+import styled from "styled-components/native";
+import Quotes from "./components/Quotes";
+import { Text } from "react-native";
```
JennieDalgren, line 4: since you're not using this you don't need to import it.  when setting up the styled components it'll sort that out itself to "find" the right element.

App.js
```diff
@@ -1,26 +1,34 @@
...
   align-items: center;
-`
+`;
 
 const Title = styled.Text`
+  font-size: 34px;
+  font-weight: bold;
+  color: #ffffff;
+`;
+const Subtitle = styled.Text`
```
JennieDalgren, line 18: super picky here but add a blank line between 17 and 18 :wink

components/Quotes.js
```diff
@@ -0,0 +1,77 @@
...
+    "Don’t move. Gonorrhea can’t see us if we don’t move...Wait! I was wrong! I was thinking of a T-rex.",
+
+    "Traditionally, science fairs are a father-son thing.Well, scientifically, traditions are an idiot thing.",
+  ];
+
+  const [quote, setQuote] = useState([]);
+
+  const getQuote = () => {
+    const theQuote = quoteArray[Math.floor(Math.random() * quoteArray.length)];
+    setQuote(theQuote);
```
JennieDalgren, line 49: nice little way of getting a random quote

components/Quotes.js
```diff
@@ -0,0 +1,77 @@
...
+
+  return (
+    <View>
+      <Container>
+        <Text style={{ fontSize: 20 }}>{quote}</Text>
+      </Container>
+      <Button
+        onPress={() => {
+          getQuote();
+          Vibration.vibrate();
```
JennieDalgren, line 60: nice that you've tried out the vibration functionality

components/Quotes.js
```diff
@@ -0,0 +1,77 @@
...
+        }}
+      >
+        <Text
+          style={{
+            color: "white",
+            fontWeight: "bold",
+            fontSize: 20,
+          }}
```
JennieDalgren, line 68: try to stay away from inline styling, make it into a styled component instead.

App.js
elinaarenstrup: Thnx, will do :)

components/Quotes.js
elinaarenstrup: Thnx for the Rick & Morty api btw. I will def use that one in the near future :D

PR #47, opened 2020-04-10
React-Native-Emelie

emlo2018:
This week I had a long startup and it took a long time for me to get in to the documentation and to use expo. I had a lot of ideas that I wanted to try out but in the end I had to keep it as simple as possible just to hand in something that works for android and ios. 


components/Detail.js
```diff
@@ -0,0 +1,47 @@
...
+import { SafeAreaView, ScrollView, StyleSheet } from 'react-native'
+
+export const Detail = () => {
+  const [info, setInfo] = useState([])
+
+  useEffect(() => {
+    fetch('https://api.nasa.gov/planetary/apod?api_key=qSlGOjPnCXc7pgWt5QcdsI3nw45FfocwP6EggaOZ')
+      .then((res) => res.json())
+      .then((json) => setInfo(json))
+  }, [])
```
JennieDalgren, line 12: it looks like this is exactly the same fetch as in the other component.
So I would suggest you do the useEffect in the parent component and pass along  or at least just saving the info that you need in the state variables instead of the whole thing.
So here,  if you only care about the explanation, just save the explanation:
```
const [info, setInfo] = useState('')
and
.then((json) => setInfo(json.explanation))
```

components/Nasa.js
```diff
@@ -0,0 +1,80 @@
...
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    marginTop: Constants.statusBarHeight,
+  },
+  scrollView: {
+    paddingHorizontal: 10,
+    backgroundColor: '#222',
+  }
+})
```
JennieDalgren, line 80: don't mix stylesheet and styled components

components/Tap.js
```diff
@@ -0,0 +1,39 @@
...
+align-items: center;
+`
+
+const Info = styled.Text`
+flex:1;
+justify-content: center;
+align-items: center;
+color: white;
+font-weight: bold;
+font-size: 45px;
```
JennieDalgren, line 38: indentation

emlo2018:
I did a lot off changes since last friday. The api fetch is only in Nasa.js and it works much better with expo. The pages are separeted now and the styling is corrected.

PR #56, opened 2020-04-13
react native app - Anne-Sophie

annesophiegendron:
My reflections on how this week's project turned out: It took me a long time to figure out how to install correctly all the dependencies, and to be able to display something on the simulator.  This is the first draft of what my small app is. I did not have enough time to implement the navigation (I worked on that on a different repo though), so I'll try to add it to this project later.
Link to the snack https://snack.expo.io/@annesophiegendron/dad's-joke-generator

Things I'd like to have clarified or explained in more detail: I had to install a other npm (axios) in order to make the API fetch. Why is it so? I just wonder why I could not use 'fetch' as usual with this API.

### DIFF
--- a/App.js
+++ b/App.js
@@ -1,19 +1,20 @@
 import React, { useState, useEffect } from 'react'
 import styled from 'styled-components/native'
-import { ActivityIndicator, Text } from 'react-native'
+import { ActivityIndicator } from 'react-native'
 import axios from 'axios'
 
 const Container = styled.View`
   flex: 1;
-  background-color: white;
+  background-color: #270F59;
   align-items: center;
   justify-content: center;
+  padding: 50px;
 `
 const Button = styled.TouchableOpacity`
-  background: black;
+  background: #F272DD;
   position: absolute;
-  top: 40;
-  padding: 10px 15px;
+  bottom: 140;
+  padding: 20px 25px;
   border-radius: 8;
 `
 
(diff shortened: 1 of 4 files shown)


App.js
```diff
@@ -1,26 +1,64 @@
...
+
+  useEffect(() => {
+    const fetchData = async () => {
+      setLoading(true)
+      const result = await axios("https://us-central1-dadsofunny.cloudfunctions.net/DadJokes/random/type/general/")
+      setJoke(`${result.data[0].setup} ${result.data[0].punchline}`)
+      setLoading(false)
+    }
+    fetchData()
+  }, [fetching])
```
JennieDalgren, line 49: No need to use axios. Just use the normal fetch that we've done before:
```
 useEffect(() => {
  setLoading(true)
   fetch("https://us-central1-dadsofunny.cloudfunctions.net/DadJokes/random/type/general/")
      .then(res => res.json())
      .then(json => {
         setJoke(`${json[0].setup} ${json[0].punchline}`)
         setLoading(false)
       }
  },[fetching])
```

App.js
```diff
@@ -1,26 +1,64 @@
...
+  text-align: center
 `
 
+const styles = StyleSheet.create({
+  baseText: {
+    color: 'white',
+    fontSize: 20
+  },
+})
+
```
JennieDalgren, line 34: stick to styled components!

PR #68, opened 2020-04-22
Week 13 Build react native app

Sickan-sys:

### DIFF
--- a/App.js
+++ b/App.js
@@ -1,26 +1,100 @@
 import React from 'react'
-import styled from 'styled-components/native'
+import { Platform, StyleSheet, Text, View, FlatList } from 'react-native'
+import Header from './components/Header'
+import InputBar from './components/InputBar'
+import TodoItem from './components/TodoItem'
 
-const Container = styled.View`
-  flex: 1;
-  background-color: papayawhip;
-  justify-content: center;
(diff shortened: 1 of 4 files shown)


App.js
```diff
@@ -1,26 +1,100 @@
...
+    todos = todos.map((todo) => {
+      if (todo.id == item.id) {
+        todo.done = !todo.done
+      }
+
+      return todo
+    })
+
+    this.setState({todos})
+  }
```
HIPPIEKICK, line 48: You have different naming here (todo vs todos), so the app is crashing... Change so that it corresponds, and it should work

components/TodoItem.js
```diff
@@ -0,0 +1,42 @@
...
+                { todoItem.title }
+            </Text>
+
+            <Button
+            title="Remove"
+            color={(todoItem.done) ? 'rgba(200, 0, 0, 0.5)' : 'rgba(255, 0, 0, 1)' }
+            onPress={() => this.props.removeTodo()}
+            />
+        </TouchableOpacity>
+    )
```
HIPPIEKICK, line 27: The remove button doesn't show. Try putting the button outside/below the TouachableOpacity

PR #69, opened 2020-04-22
React Native App – Kajsa Holmström

crushoftheyear:
A simple To-Do app created with React Native.

The app is split up in three components, each of those using Styled Components. Adding a new item is handled with State. Items are added to a FlatList & being removed on click.

**If I had more time:**

- Clear input after submit
- Add specific delete button to each item
- When item is clicked, different styling should be applied to indicate it's been completed

### DIFF
--- a/components/AddItem.js
+++ b/components/AddItem.js
@@ -22,9 +22,15 @@ const ButtonText = styled.Text`
 
 
 export const AddItem = ({ setTodos }) => {
+
   const [text, setText] = useState('')
 
-  // Add item
+  // Set text to be equal to input value
+  const changeHandler = (val) => {
+    setText(val)
+  }
+
+  // Add todo item
   const submitHandler = (text) => {
     setTodos((prevTodos) => {
       return [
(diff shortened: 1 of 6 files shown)


components/AddItem.js
```diff
@@ -0,0 +1,57 @@
...
+  // Set text to be equal to input value
+  const changeHandler = (val) => {
+    setText(val)
+  }
+
+  // Add todo item
+  const submitHandler = (text) => {
+    setTodos((prevTodos) => {
+      return [
+        { text: text, key: Math.random().toString() },
```
HIPPIEKICK, line 37: could also use `Date.now()`

components/TodoItem.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import styled from 'styled-components/native'
+
+export const TodoItem = ({ index, item, pressHandler }) => {
+
+  const Item = styled.TouchableOpacity``
+  const ItemText = styled.Text`
+  padding: 20px;
+  background-color: ${index(item) % 2 === 0 ? '#f8b400' : '#e3a500'};
```
HIPPIEKICK, line 9: hacky ⭐

HIPPIEKICK:
PS2. I like your color scheme 👯
Skärmavbild 2020-06-15 kl  18 14 41

PR #72, opened 2020-06-08
React Native App Åsa Andersson

aasa72:
My reflections on how this week's project turned out: Massive problems with technology, expo and I don´t get along at all!


App.js
```diff
@@ -1,26 +1,68 @@
...
+  ]);
+
+  const pressHandler = (key) => {
+    setTodo((pastTodo) => {
+      return pastTodo.filter (todo => todo.key != key);
+    });
+  };
+
+  const submitHandler = (text) => {
+    setTodo = ((pastTodo) => {
```
HIPPIEKICK, line 42: You should `setState` like this:
`setTodo((pastTodo) => {`

components/AddToDos.js
```diff
@@ -0,0 +1,49 @@
...
+        onChange={ e => setText(e.target.value)}
+        value={text}></Input>
+      
+      <AddButton
+        onPress={() => submitHandler(text)} >
+        <Title>ADD</Title>
+      </AddButton>
+    </Container>
+  );
+};
```
HIPPIEKICK, line 49: To make the text update, do something like this so that it's updating the text before sending it further. I got the text as undefined now...

```
export const AddToDos = ({ submitHandler }) => {
  const [ text, setText ] = useState('');

  const changeText = (value) => {
    setText(value)
  }

  return (
    <Container>
      <Input 
        placeholder='What to do...'
        type="text"
        onChangeText={changeText}
        value={text}/>
      <AddButton
        onPress={() => {submitHandler(text)}} >
        <Title>ADD</Title>
      </AddButton>
    </Container>
  );
};
```

PR #74, opened 2020-11-10
Magic Oracle 🔮 - Sofia Vaz Sousa

sofiavazs:
I'm really happy on how this app turned out,  I managed to reach the blue level requirements and some stretch goals. :)

Debugging in general, I feel that's something I struggled with, so I used my phone a lot and for some reason.. since adding the Lottie animations that I wasn't able to check the web version.. so maybe I need some clarification on that


App.js
```diff
@@ -1,26 +1,48 @@
...
-  color: palevioletred;
-`
+import { OracleMessage } from './Components/OracleMessage';
+import { ShakeEventFunction } from './Components/ShakeEventFunction';
+import { StartScreen } from './Components/StartScreen';
+import { AboutApp } from './Components/AboutApp';
 
 const App = () => {
+  //state that stores the state of the initial screen
+  const [initialScreen, setInitialScreen] = useState(true)
```
EmOlsz, line 13: It seems that you sometimes add semicolons, and sometimes don't - let's be consistent about that :)

Components/OracleMessage.js
```diff
@@ -0,0 +1,31 @@
+import React, { useEffect, useState } from 'react';
+import LottieView from 'lottie-react-native';
+import { Platform } from 'react-native';
+
+import { Container, OracleText, MagicImage, ButtonText, RestartButton } from '../StyledComponents/OracleMessageStyling';
```
EmOlsz, line 5: No need to name your folder with capital letter, this convention is saved only for :

1. Names of components
2. Names of files of components

Components/OracleMessage.js
```diff
@@ -0,0 +1,31 @@
...
+  useEffect (()=> { 
+    fetch('https://api.adviceslip.com/advice')
+      .then (response => response.json())
+      .then ((json) => { 
+        setMessage(json.slip) 
+      });
+  },[]);
+
+  return (
+    <>
```
EmOlsz, line 19: No need for React Fragment as you only return one element (`<Container />`)

App.js
```diff
@@ -1,26 +1,48 @@
...
+  const shake = () => {
+    setInitialScreen(false);
+  };
+
+  useEffect(() => {
+    ShakeEventFunction.addListener(() => {
+      shake();
+    })
+    return () => {
+      ShakeEventFunction.removeListener();
```
EmOlsz, line 26: Great job - it's very important to remember about clean up on component unmount (small history extra - before hooks, we did that by using special lifecycle method called `componentWillUnmount()`)

EmOlsz (approved):
Congratulations Sofia, great work this week! Not so many things I could point out :)
Nonetheless, take a look on my comments, hope they might help you for the future!

PR #76, opened 2020-11-12
The world's best dad jokes - Annika Melin

ankimelin:
I'm happy I built something on a simple API, as I quite quickly had a base application to continue on. I wanted to practise text input, and it was fun to implement a share function. Tried out "try/catch" and "finally" for the first time.

### DIFF
--- a/App.js
+++ b/App.js
@@ -71,7 +71,7 @@ const ButtonContainer = styled.View`
 const App = () => {
 
   const [joke, setJoke] = useState("")
-  const [search, setSearch] = useState()
+  const [searchWord, setSearchWord] = useState()
 
   const getJoke = () => {
     fetch('https://icanhazdadjoke.com/', {
@@ -84,8 +84,8 @@ const App = () => {
       .then(json => setJoke(json.joke))
   }
 
-  const searchJoke = () => {
-    fetch(`https://icanhazdadjoke.com/search?term=${search}`, {
+  const getSearchedJoke = () => {
+    fetch(`https://icanhazdadjoke.com/search?term=${searchWord}`, {
       method: 'GET',
       headers: {
         "Accept": "application/json"
@@ -100,8 +100,8 @@ const App = () => {
         alert("No such joke")
       })
       .finally(() => {
-        setSearch("")
-        setSearch()
+        setSearchWord("") // empties the textinput, without it tomato stays but also no such joke is displayed
+        setSearchWord() // empties the state, wihtout it you can search new joke on empty string
       })
   }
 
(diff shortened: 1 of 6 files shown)


App.js
```diff
@@ -1,26 +1,151 @@
...
+      method: 'GET',
+      headers: {
+        "Accept": "application/json"
+      }
+    })
+      .then(response => response.json())
+      .then(json => setJoke(json.joke))
+  }
+
+  const searchJoke = () => {
```
EmOlsz, line 87: From naming point of view, `getJoke()` and `searchJoke()` functions are not very descriptive. I would rather suggest something like `getJoke()` and `getSearchedJoke()`

App.js
```diff
@@ -1,26 +1,151 @@
...
+      })
+  }
+
+  const shareJoke = async () => {
+    try {
+      const result = await Share.share({
+        message:
+          `${joke}`,
+      })
+      result.action === Share.sharedAction ? (alert("Joke shared")) :
```
EmOlsz, line 114: No need to wrap `alert()` inside extra set of parentheses, this is completely enough :

```
result.action === Share.sharedAction ? alert("Joke shared") : 
```

App.js
```diff
@@ -1,26 +1,151 @@
...
+      .catch(() => {
+        alert("No such joke")
+      })
+      .finally(() => {
+        setSearch("")
+        setSearch()
+      })
+  }
+
+  const shareJoke = async () => {
```
EmOlsz, line 108: Great job with `async await`:)
For further explanation of this topic, check out https://medium.com/javascript-in-plain-english/async-await-javascript-5038668ec6eb

App.js
```diff
@@ -1,26 +1,151 @@
...
+      .then(json => {
+        const randomIndex = Math.floor(Math.random() * Math.floor(json.results.length))
+        setJoke(json.results[randomIndex].joke)
+      })
+      .catch(() => {
+        alert("No such joke")
+      })
+      .finally(() => {
+        setSearch("")
+        setSearch()
```
EmOlsz, line 104: I think you don't need one of those lines

EmOlsz (changes requested):

App.js
ankimelin: makes sense, updated! 

also updated search and setSearch to searchWord and setSearchWord

App.js
ankimelin: fixed!

App.js
ankimelin: thanks!

App.js
ankimelin: Actually I do, setSearch("") (now setSearchWord("")) will empty the textinput field, but the state will hold and empty string which is totally fine to query jokes on from the api - I don't want that. So what setSearch() (not setSearchWord()) does is to empty the state, with this one alone the textinput field is not emptied - only the state is emptied.

ankimelin:
I have updated/answered on the comments requesting changes, please let me know if the project is goot to go!

EmOlsz (approved):

PR #79, opened 2020-11-12
Space Is Cool!

jonnaru:
React Native made me depressed. I've been shaking my phone to reload the whole week. Please let me pass!

No thank you, brain is overloaded 😖

**View it live:**
https://snack.expo.io/@jonnaru/space-is-cool


components/ChangeImage.js
```diff
@@ -0,0 +1,42 @@
...
+
+const ChangeImage = ({
+  text,
+  setImgNumber,
+  setQuery,
+  query,
+  newQuery,
+  random,
+}) => {
+  const handleOnPress = () => {
```
EmOlsz, line 27: Since you use this function in the same component you declare it in, it shouldn't have `handle` in the name. `onButtonPress` or `onButtonClick` would be more appropriate

EmOlsz (approved):

PR #84, opened 2020-11-13
APP: Workout of the day - Gabriella Bolin

Gabbi-89:
It was extremely hard to know where to start, then we I add something to show it was easier to build thing upon that.


App.js
```diff
@@ -1,26 +1,23 @@
-import React from 'react'
-import styled from 'styled-components/native'
+import React from 'react';
+import { NavigationContainer } from '@react-navigation/native';
+import { createStackNavigator} from '@react-navigation/stack';
```
HIPPIEKICK, line 3: Nice with navigation 👌

components/Header.js
```diff
@@ -0,0 +1,27 @@
...
+
+  if (!fontLoaded) {
+    return <AppLoading />
+  } else {
+    return <Title>JUST DO IT!</Title>
+  };
+};
+
+const Title = styled.Text`
+  font-family: 'Inter-Regular';
```
HIPPIEKICK, line 19: Hmm, I still get an error that this is not a system font. Maybe add a backup font?

components/WorkoutDetail.js
```diff
@@ -0,0 +1,65 @@
...
+  });
+  
+  return (
+    <Container>
+      <GifImage source={{uri: 'https://media4.giphy.com/media/tn8zWeNYA73G0/200.webp?cid=ecf05e47x13nr03fqr4i6gvj673lsj1w6nim7jgp9ayiopsi&rid=200.webp'}} />
+      <DoTraining>What are you waiting for?</DoTraining>
+      <DoTraining>Just go away {workout.type}!</DoTraining>
+      <BackButton onPress={() => navigation.goBack()}>
+        <TextButton>Go Back!</TextButton>
+      </BackButton>
```
HIPPIEKICK, line 18: Honestly, I didn't understand the images were clickable until I read this row 😅 Is it possible to make it clearer to the user that it's possible to click the advice on the first screen?

components/WorkoutType.js
```diff
@@ -0,0 +1,118 @@
...
+        color: '#5F91A3',
+        image: require('../assets/weightlift.png')
+      },
+      {
+        type: 'Walk', 
+        color: '#5FA390',
+        image: require('../assets/walking.png')
+      },
+      {
+        type: 'Rest Day',
```
HIPPIEKICK, line 32: Minor thing, but later when you concatenate the string - it adds up to "Just go away rest day" - which _do_ rhyme but it's good to think about if all data fits in the string your adding to :)

components/WorkoutDetail.js
HIPPIEKICK: Also, it would be nice if this screen were scrollable since the content doesn't always fit (on my phone at least) - so the back button gets so far down it's almost not clickable. Try giving the button some bottom padding

components/WorkoutType.js
HIPPIEKICK: Also, this array could be located in its own file and imported into this file - for some more structure in your project :)

HIPPIEKICK (approved):
Gold star for perfectly indented Styled Components ⭐ Nice usage of navigation and good structure in your project. Good component names - keep it up!

PR #85, opened 2020-11-13
Hogwarts sorting ceremony ⚡ React Native ⚡ Andrea

Andrea-Osmar:
My reflections on how this week's project turned out: I am very happy that I have a finished project! I could have done more but I'm just glas that it doesn't come back with lots of errors.

Things I'd like to have clarified or explained in more detail: I think for next year it would have been great with option of following a set design. Would also have been helpful if you would to post som codealong material. I did not find any codealong tutorials that I found relevant. 
I would like to know how to publish this if I want to show my family.


App.js
```diff
@@ -1,26 +1,45 @@
-import React from 'react'
-import styled from 'styled-components/native'
+import React, { useState, useEffect } from "react";
 
-const Container = styled.View`
+import { Firstpage } from "./components/Firstpage";
```
HIPPIEKICK, line 3: You need to be consistent in your naming, the file is called `FirstPage.js` but everywhere else you call it `Firstpage`. PascalCase is the way to go 👍

App.js
```diff
@@ -1,26 +1,45 @@
...
+import { Houses } from "./components/Houses";
+import { ShakeEvent } from "./components/ShakeEvent";
```
HIPPIEKICK, line 5: Fun that you're using the shake event!

components/Houses.js
```diff
@@ -0,0 +1,61 @@
...
+export const Houses = ({ onRestart }) => {
+  const [house, setHouse] = useState();
+  const [image, setImage] = useState();
+
+  useEffect(() => {
+    fetch('https://www.potterapi.com/v1/sortinghat')
+      .then((result) => result.json())
+      .then((data) => {
+        setHouse(data);
+        setImage(data.toLowerCase());
```
HIPPIEKICK, line 42: Do you really need two states?

components/Houses.js
```diff
@@ -0,0 +1,61 @@
...
+        vibrateDevice();
+      });
+  }, []);
+
+
+  const vibrateDevice = () => {
+    Vibration.vibrate();
+  }
```
HIPPIEKICK, line 50: Nice!

assets/images.js
```diff
@@ -0,0 +1,7 @@
+const images = {
+  gryffindor: require("./gryffindor.png"),
+  hufflepuff: require("./hufflepuff.png"),
+  slytherin: require("./slytherin.png"),
+  ravenclaw: require("./ravenclaw.png"),
+};
+export default images;
```
HIPPIEKICK, line 7: So nice that this got its own file 💯

HIPPIEKICK (approved):
Nice job Andrea - just be careful with upper and lower case characters (it can mess things up). Maybe the house-component should only be called `House.js` since it's only one house at a time showing?

HIPPIEKICK:
Oh, and a link on how to publish -> https://instabug.com/blog/react-native-app-ios-android/

PR #86, opened 2020-11-13
Kanye-Generator by Anna Hasselström

annahasselstrom:
It was tricky to take my knowledge of js, css, and html and put it to use in React Native. I had lots of warnings and errors along the way. Also building for different devices was hard. But I also really enjoyed it. There is something very satisfying in viewing and interacting with your results on a mobile. I can definitely see myself learning more about React Native - and having fun doing it.

Main learnings: I need to become more structured and methodological when debugging - both to same time and sanity 😀


App.js
```diff
@@ -1,25 +1,27 @@
-import React from 'react'
-import styled from 'styled-components/native'
+import React from 'react';
+import { NavigationContainer } from "@react-navigation/native";
+import { createStackNavigator } from "@react-navigation/stack";
```
HIPPIEKICK, line 3: Super nice navigation! ⭐

components/QuoteList.js
```diff
@@ -0,0 +1,74 @@
...
+
+export default function QuoteList() {
+  const [newQuote, setNewQuote] = useState('');
+  const [refreshing, setRefreshing] = React.useState(false);
+
+  // Function to fetch the quote
+  const fetchQuote = () => {
+    fetch('https://api.kanye.rest')
+      .then((res) => res.json())
+      .then((res) => {
```
HIPPIEKICK, line 34:
```
.then(res => res.json())
.then(res => {
```

components/QuoteList.js
```diff
@@ -0,0 +1,74 @@
...
+      .catch((error) => console.error(error));
+  };
+
+  // When the pulldown is used, we:
+  //  - show the refresh animation
+  //  - fetch the new quote
+  const onRefresh = React.useCallback(() => {
+    setRefreshing(true); // but when true go ahead and refresh (and generate new quote).
+    fetchQuote();
+  });
```
HIPPIEKICK, line 50: I really like this idea!

components/QuoteList.js
```diff
@@ -0,0 +1,74 @@
...
+  // When newQuote changes, we want to do this, hence the dependency
+
+  useEffect(() => {
+    if(newQuote !=('')){
```
HIPPIEKICK, line 54: Consider `if(!newQuote)`

components/QuoteList.js
```diff
@@ -0,0 +1,74 @@
...
+      Alert.alert(JSON.stringify(newQuote));
+    }
+  }, [newQuote]);
+  
+  return (
+    <SafeAreaView>
+      <ScrollView
+        refreshControl={
+          <RefreshControl refreshing={refreshing} onRefresh={onRefresh} />
```
HIPPIEKICK, line 63: Some new lines for attributes would look 👍

components/QuoteList.js
```diff
@@ -0,0 +1,74 @@
...
+        }
+      >
+        <TopContainer>
+          <Title>Kanye West quotes</Title>
+          <Title>Pull me down!</Title>
+          <Image source={kanye} />
+        </TopContainer>
+      </ScrollView>
+    </SafeAreaView>
+  )
```
HIPPIEKICK, line 73: ;

components/QuoteList.js
```diff
@@ -0,0 +1,74 @@
...
+  color: white;
+`
+
+const Image = styled.Image`
+  width: 150px;
+  height: 150px;
+  margin-top: 20px;
+`    
+
+export default function QuoteList() {
```
HIPPIEKICK, line 26: Why not arrow function? Be consistent!

HIPPIEKICK (approved):
Wow Anna, you managed to do a lot! Navigation, control the refresh, API 👌 Consider renaming the `QuoteList.js` component since it only returns one quote at a time.

PR #88, opened 2020-11-13
What to do on a rainy day ☔️

Snabbina:
The det up and all the dependencies was a project on its own, otherwise cool to make a working mobile app in just a few days!

### DIFF
--- a/App.js
+++ b/App.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import styled from 'styled-components/native'
+import { ThingsToDo } from './components/ThingsToDo'
 
 const Container = styled.View`
   flex: 1;
@@ -16,9 +17,11 @@ const Title = styled.Text`
 const App = () => {
   return (
     <Container>
-      <Title>This is your cool app!</Title>
-      <Title>Go to App.js and start coding</Title>
-      <Title>💅💅💅</Title>
+      {/* <Title>This is your cool app!</Title>
+      <Title>Go to App.js and start coding</Title> */}
+      <Title>Vad ska vi göra idag Nathaniel?</Title>
+      <ThingsToDo />
+
     </Container>
   )
 }
--- a/App.js
+++ b/App.js
@@ -1,28 +1,28 @@
-import React from 'react'
-import styled from 'styled-components/native'
-import { ThingsToDo } from './components/ThingsToDo'
+import React from "react";
+// import styled from 'styled-components/native'
+
+import { ThingsToDo } from "./components/ThingsToDo";
 
 const Container = styled.View`
   flex: 2;
   background-color: #c0d6e4;
   justify-content: center;
   align-items: center;
-`
+`;
 
 const Title = styled.Text`
   font-size: 30px;
   color: black;
   margin: 20px;
   text-align: center;
-`
-
+`;
 const App = () => {
   return (
     <Container>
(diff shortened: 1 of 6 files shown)


App.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
 import styled from 'styled-components/native'
+import { ThingsToDo } from './components/ThingsToDo'
```
EmOlsz, line 3: Let's not forget about one empty line between outer and local dependencies!

components/ThingsToDo.js
```diff
@@ -0,0 +1,98 @@
+import React, { useState } from "react";
+import styled from "styled-components/native";
+import { Text, View, Vibration, TouchableOpacity } from 'react-native';
```
EmOlsz, line 3: I see that sometimes you use single quotes and sometimes double quotes - let's try to be consistent about it

components/ThingsToDo.js
```diff
@@ -0,0 +1,98 @@
...
+const ThingsToDoText = styled.Text`
+  font-size: 20px;
+  text-align: center;
+  color: white;
+`;
+
+export const ThingsToDo = () => {
+  // array of objects of things to do
+
+  const ThingsToDoArray = [
```
EmOlsz, line 44: Uppercase names are reserved for components only, not for arrays

components/ThingsToDo.js
```diff
@@ -0,0 +1,98 @@
...
+      {
+        thingtodo: "Ha på pyamas hela dagen!",
+        color: "#e2dcd4", //grey
+        emoji: "",
+      },
+  ];
+
+  // An randommaking function that gets invoked when button gets push.
+
+    const [thingtodo, setThingToDo] = useState({});
```
EmOlsz, line 74: Wrong indentation over here ;)

components/ThingsToDo.js
```diff
@@ -0,0 +1,98 @@
...
+
+  const getThingsToDo = () => {
+    const theThingToDo =
+      ThingsToDoArray[Math.floor(Math.random() * ThingsToDoArray.length)];
+    setThingToDo(theThingToDo);
+  };
+
+  return (
+    <Container>
+      <ThingsToDoContainer style={{ backgroundColor: thingtodo.color }}>
```
EmOlsz, line 84: If we decide to use styled components, then let's stick to it and do not pass styles via `style` attribute.
You can pass `thingtodo` value as prop to your styled component and later on use it there to define styles

components/ThingsToDo.js
```diff
@@ -0,0 +1,98 @@
...
+        <ThingsToDoText>{thingtodo.thingtodo}</ThingsToDoText>
+        <ThingsToDoText>{thingtodo.emoji}</ThingsToDoText>
+      </ThingsToDoContainer>
+      <PickButton
+        onPress={() => {
+          getThingsToDo();
+          Vibration.vibrate();
```
EmOlsz, line 91: If we need to execute more than one line of code, then we should move our logic to separate function and call it on `onPress` :

```
const myCustomFunc = () => {
  getThingsToDo();
  Vibration.vibrate();
}

return (
  <PickButton onPress={myCustomFunc} />
)
```

EmOlsz (changes requested):

Snabbina:
Hello, made the changes requested on this project, is it ok now? /S

EmOlsz (approved):

PR #89, opened 2020-11-13
Letter App 

moablomkvist:
It was fun and challenging. I cant wait for when I have the time to build more functions to this app. 

https://expo.io/@moablomkvist/projects/project-react-native-app


App.js
```diff
@@ -1,26 +1,28 @@
 import React from 'react'
+import data from './data.json'
+
+import { Alfabet } from './components/Alfabet.js'
 import styled from 'styled-components/native'
```
EmOlsz, line 5: `styled-components` is an outer dependency, therefore it should be placed next to `react`

App.js
```diff
@@ -1,26 +1,28 @@
...
-      <Title>This is your cool app!</Title>
-      <Title>Go to App.js and start coding</Title>
-      <Title>💅💅💅</Title>
-    </Container>
-  )
+
+return (
+  <Container>
+    <Alfabet 
+    letterArray={data.alfabet}
```
EmOlsz, line 22: Some missing indentation

components/Alfabet.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import { useState } from 'react'
+import { View, Vibration } from 'react-native'
+
+import styled from 'styled-components/native'
```
EmOlsz, line 5: Once again, this package shouldn't be separated from the others by one empty line - they are all outer packages, so they should stay together

components/Alfabet.js
```diff
@@ -0,0 +1,62 @@
...
+
+const TouchButton = styled.TouchableOpacity`
+  padding: 25px;
+  margin: 0px;
+  width: 250px;
+  borderRadius: 25px;
```
EmOlsz, line 11: We shouldn't use camelCase in styled components, normal kebab-case is ok. So :

```
border-radius
```

instead of 

```
borderRadius
```

components/Alfabet.js
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <View>
+        <Emoji>
+          {emoji}
+        </Emoji>
+        <Letter>
+          {letter}
+        </Letter>  
+        <TouchButton onPress={() => { 
+        getLetter(); Vibration.vibrate()}}>
```
EmOlsz, line 56: If we need to execute more than one line of code, then we should move our logic to separate function and call it on onPress :
```
const myCustomFunc = () => {
  getLetter();
  Vibration.vibrate();
}

return (
  <TouchButton onPress={myCustomFunc} />
)
```

EmOlsz (changes requested):

EmOlsz:
Thank you for changes Moa, project approved!

EmOlsz (approved):

PR #90, opened 2020-11-13
StopWatch App-Tara

trangtara:


App.js
```diff
@@ -1,26 +1,191 @@
...
+              }
+              return seconds + 1;
+            });
+          }
+          return hundredths + 1;
+        });
+      }, 10);
+      return () => clearInterval(interval);
+    }
+  }, [isRunning]);
```
EmOlsz, line 32: No need to trigger this `useEffect()` each time `isRunning` is changed, interval once initialised will know to execute itself every 10ms

App.js
```diff
@@ -1,26 +1,191 @@
...
+  const handleStop = () => {
+    setIsRunning(false);
+  };
+
+  const handleLap = () => {
+    const lap = {
+      num: laps.length + 1,
+      time: formattedTime
+    };
+    setLaps((laps) => [lap, ...laps]);
```
EmOlsz, line 54: Great job with state argument :)

App.js
```diff
@@ -1,26 +1,191 @@
...
+              </LapButton>
+            }
+          </LapResetContainer>
+        </ButtonContainer>
+      </FixedSection>
+      <ScrollView>
+        <LapContainer>
+          {laps.map((item) => (
+            <LapListItem 
+              key={item.num}>
```
EmOlsz, line 97: No need for line break here, you only pass one prop to `<LapListItem />` component

App.js
```diff
@@ -1,26 +1,191 @@
...
+              <Span>Lap {item.num}</Span><Span>{item.time}</Span>
```
EmOlsz, line 98: If it's styled `<span />`, we should rather named it with some informative name, not only with capitalised letter

EmOlsz (approved):

PR #96, opened 2020-11-14
React Native app / Katarina 

Katsiez:
This week has been the most difficult one so far. 

I felt like we've gotten more than enough guidance and help this week so everything was great from that standpoint. I do wish I had had it in me to fetch the actual data from the API but I just felt like I needed a break after this week.


EmOlsz (changes requested):
Hey Katarina, nice project, I'm impressed! You've implemented some complex functionality!
Nonetheless, it's obvious that this week you used code snippets you found somewhere else. It's not bad at all, let me make it straight. Even more, it's very good that you take an advantage of using code done by other developers :)
However, I would like you to remember about modifying this kind of code next time. In 99% of cases we can (and even should) adapt copied code to our local context - in your case it would be rearranging class-based components to functional components with hooks, changing class-based lifecycle methods to functional useEffect etc. Thanks to that, our code won't be chaotic and it will be very easy to maintain it in longer period.
Apart from that, good job!

For now, let's add styled components to your project, after that I'll be able to approve pull request :)

HIPPIEKICK (approved):
Nice quiz! ⭐

PR #97, opened 2020-11-14
Rock-paper-scissors-lizard-spock - Anna Hellqvist

AnnaHellqvist:
I'm kind of happy with how the project turned out but it has been a struggle this week. 
I didn't really have time to fix all the things I wanted to, I wanted to work a bit more on my styling, add a global styling file. And reused the styling more since I now have almost the same styling in every screen.. 
I had some trouble to get the layout to work out as I wanted it to.
But I think I will try to play around a bit more in react native when the bootcamp is over, it seems fun when you get the hang of it. 

I tried to pass the userPick state as an object with a string and the variable/filepath to image but that didn't work out. It kept complaing about that it was an object. It seems to work with my compPick state so I do not get why I couldnt do the same with the userPick... I solved it by creating two states instead, one with a string and one with the variable/filepath.


App.js
```diff
@@ -1,25 +1,35 @@
...
-      <Title>This is your cool app!</Title>
-      <Title>Go to App.js and start coding</Title>
-      <Title>💅💅💅</Title>
-    </Container>
+    <NavigationContainer>
+      <Stack.Navigator>
+        <Stack.Screen
+          name="Start"
+          component={StartScreen}
+          options={{ headerShown: false }}
```
EmOlsz, line 19: Great job with hiding header :)

components/UserPickScreen.js
```diff
@@ -0,0 +1,161 @@
...
+    setUserImage(path);
+    // Randomizes a choice for the computer from the choices array
+    setCompPick(choices[Math.floor(Math.random() * choices.length)]);
+  }
+
+  // Navigates to the next page and sends props
+  const navigateToResult = () => {
+    navigation.navigate("Result", {
+      compChoice: compPick,
+      userChoice: userPick,
```
EmOlsz, line 111: Or we could simplify this by unifying names in local state :

```
navigation.navigate("Result", { compChoice, userChoice, userImage });
```

EmOlsz (approved):

PR #98, opened 2020-11-14
React Native Emelie 

emeliesv:
I'm glad that I managed to pull this off! 
I created a multipage app that uses an API and also sends props of data from one component to another :)
Also, this serves a personal purpose as well since I often don't know what to have for dinner!


components/Container.js
```diff
@@ -0,0 +1,32 @@
+
+import styled from 'styled-components/native'
+
+export const Container = styled.ImageBackground`
```
EmOlsz, line 4: Great idea to store reusable components in separate file.
Not sure though why this file is named `Container.js`, it has multiple components inside, not only `<Container />`

components/Footer.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import styled from 'styled-components/native'
+
+export const Footer = () => {
+    const Container = styled.View`
```
EmOlsz, line 5: Styled components should be declared outside of any component (ideally, below declaration of `<Footer />`)

components/Header.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import styled from 'styled-components/native'
+
+export const Header = () => {
+
+const Container = styled.View`
+flex: 0.2;
```
EmOlsz, line 7: Wrong indentations, over here and in couple of different places

components/HomeScreen.js
```diff
@@ -0,0 +1,28 @@
...
+font-size: 48px;
+`
+
+export const HomeScreen = ({navigation}) => {
+    useEffect(() => {
+        navigation.setOptions({headerShown:false})
+    }, [])
+
+    const navigateToRecipe = () => {
+        console.log('This works!')
```
EmOlsz, line 18: We should delete all console.logs before final push

components/HomeScreen.js
```diff
@@ -0,0 +1,28 @@
...
+        navigation.navigate('Recipes')
+    }
+
+    return (
+        <Container source={background}>
+            <HomeText>What's for dinner?</HomeText>
+            <Button title={'Find out!'} onPress={navigateToRecipe}></Button>
```
EmOlsz, line 25: We don't need to wrap string with curly brackets (`{}`)

```
title="Find out!"
```

components/Recipe.js
```diff
@@ -0,0 +1,71 @@
...
+        getRandomRecipes()
+    }, [])
+
+    return (
+        <Wrapper>
+            {recipes.map((recipe) => (
+                <Button
+                    key={recipe.id}
+                    onPress={() => navigation.navigate('Recipe-details', {
+                        heading: recipe.title,
```
EmOlsz, line 58: If we need to write more than one line of logic inside event handler, then it should be moved to separate function and called inside event handler

EmOlsz (changes requested):

EmOlsz (approved):

PR #99, opened 2020-11-15
React Native, Swdish RPG character randomizer

LillanEriksson:

I had a rough week, after making my app work out just fine, the API I had chosen to work with stopped working. So I had to come up with a new idea two days before deadline. I had some code I wrote for another personal project that I re-used for this app, combining with a API for random names.
It was a fun week, now that I'm finally finished. And really cool to realize that I made something comletely new and in a new language in just a week!

### DIFF
--- a/App.js
+++ b/App.js
@@ -1,26 +1,70 @@
-import React from 'react'
-import styled from 'styled-components/native'
+import React, { useState } from 'react';
+import styled from 'styled-components/native';
+
+import CharacterRandomizer from './CharacterRandomizer';
 
 const Container = styled.View`
-  flex: 1;
-  background-color: papayawhip;
-  justify-content: center;
-  align-items: center;
-`
+	flex: 1;
+	background-color: #443423;
+	justify-content: center;
+	align-items: center;
+`;
 
 const Title = styled.Text`
-  font-size: 24px;
-  color: palevioletred;
-`
+	font-size: 24px;
+	color: #dbc3a0;
+	margin-top: 50px;
+`;
+
+const StyledImage = styled.Image`
+	width: 300px;
+	height: 300px;
+	margin: 10px;
+	border-radius: 12px;
+`;
+
+const StyledButton = styled.Button`
+	font-size: 24px;
+	color: palevioletred;
+	background-color: #dbc3a0;
+	margin: 10px;
+`;
+
+const image = { uri: 'https://picsum.photos/id/1028/200/300' };
 
 const App = () => {
-  return (
-    <Container>
-      <Title>This is your cool app!</Title>
-      <Title>Go to App.js and start coding</Title>
-      <Title>💅💅💅</Title>
-    </Container>
-  )
-}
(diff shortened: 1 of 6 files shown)


App.js
```diff
@@ -1,26 +1,70 @@
...
-      <Title>This is your cool app!</Title>
-      <Title>Go to App.js and start coding</Title>
-      <Title>💅💅💅</Title>
-    </Container>
-  )
-}
-
-export default App
+	const [showRandomCharacter, setShowRandomCharacter] = useState(false);
+	const [hideWelcomeScreen, setHideWelcomeScreen] = useState(false);
```
EmOlsz, line 37: From naming point of view, those two state properties shouldn't have possible state value in their names (show, hide), but rather information about what they do. So instead of :

```
[showRandomCharacter, setShowRandomCharacter]
[hideWelcomeScreen, setHideWelcomeScreen]
```

they should rather be named :

```
[randomCharacterVisibility, setRandomCharacterVisibility]
[welcomeScreenVisibility, setWelcomeScreenVisibility]
```

App.js
```diff
@@ -1,26 +1,70 @@
...
+		setShowRandomCharacter(false);
+		setHideWelcomeScreen(false);
+	};
+	return (
+		<Container>
+			{!hideWelcomeScreen && (
+				<Container>
+					<Title>Swedish RPG character-randomizer</Title>
+					<StyledImage source={image} />
+					<StyledButton title={'Start your adventure!'} onPress={handlePress} />
```
EmOlsz, line 54: We don't have to wrap string values inside curly brackets (`{}`) :

```
title="Start your adventure!"
```

CharacterRandomizer.js
```diff
@@ -0,0 +1,112 @@
+import React, { useState, useEffect } from 'react';
+import { Text, View } from 'react-native';
+import styled from 'styled-components/native';
+
+//Note! Some comments are in Swedish, beacuse this app is mainly focused on swedish rpg.
```
EmOlsz, line 5: Thank you for that comment, but it doesn't change anything - all comments and instructions should be written in English. We cannot predict who will use this code later on, it should be accessible to all developers, not only Swedish speakers

CharacterRandomizer.js
```diff
@@ -0,0 +1,112 @@
...
+	//can't use class because it's a reserved keyword
+	const klass = ['Warrior', 'Magician', 'Thief'];
+
+	const randomKlass = () => randomSelector(klass);
+	//ödesmakt
+	const faith = ['Order', 'Chaos', 'Balance'];
+
+	const randomFaith = () => randomSelector(faith);
+	//styrka
+	const strenght = Math.floor(Math.random() * 18) + 3;
```
EmOlsz, line 53: `Math.floor(Math.random() * 18) + 3;` should be moved to separate function since we use it multiple times

EmOlsz (changes requested):

EmOlsz (approved):

PR #101, opened 2020-11-15
React Native app - Ingela Löfgren

IngelaL:
My reflections on how this week's project turned out: It was difficult and I am disappointed it did'nt work to share an image of a dog with a friend, just a message. So it did'nt turn out the way I wanted.

I think it would have been better if we were given a specific task to do this week. Only to set up everything took almost one day.


App.js
```diff
@@ -1,24 +1,27 @@
-import React from 'react'
+import React, { useState } from 'react'
 import styled from 'styled-components/native'
+import background from './assets/background.jpg'
```
EmOlsz, line 3: Next time let's remember about line empty line space between outer and local dependencies

components/RandomDog.js
```diff
@@ -0,0 +1,57 @@
...
+    ` 
+
+export const RandomDog = () => {
+    const [dogs, setDogs] = useState({})      
+    
+    const fetchDogs = () => {
+        fetch('https://dog.ceo/api/breeds/image/random')
+        .then((response) => response.json())
+        .then((json) => { 
+            console.log(json)
```
EmOlsz, line 34: We should remove all console.logs before final push

components/RandomDog.js
```diff
@@ -0,0 +1,57 @@
...
+            setDogs(json)
+        })
+     }   
+     /* Here we call the function so that a dog is shown when the page is mounted */
+        useEffect(() => {
+            fetchDogs()
+        }, [] )
+    
+         console.log(dogs)
```
EmOlsz, line 43: And here as well

components/RandomDog.js
```diff
@@ -0,0 +1,57 @@
...
+     /* Here we call the function so that a dog is shown when the page is mounted */
+        useEffect(() => {
+            fetchDogs()
+        }, [] )
+    
+         console.log(dogs)   
+        return (
+            <>
+                <DogView>
+                    <Image source={{uri: dogs.message}} style= {{width: 270, height: 270}}/>
```
EmOlsz, line 47: You should rather pass those styles (`width` and `height`) as in styled component

components/ShareDog.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import styled from 'styled-components/native'
+import { Share, View, Button } from 'react-native'
+
+
+
+export const ShareDog = () => {
+  const onShare = async () => {
```
EmOlsz, line 8: Congratulations about `async await` usage :)
For further explanation of this topic you can check out https://medium.com/javascript-in-plain-english/async-await-javascript-5038668ec6eb

components/ShareDog.js
```diff
@@ -0,0 +1,32 @@
...
+    try {
+      const result = await Share.share({
+        message:
+          'Here you will find the API used in the Dog randomizer: https://dog.ceo/dog-api/',
+      });
+      if (result.action === Share.sharedAction) {
+        if (result.activityType) {
+          
+        } else {
```
EmOlsz, line 17: I don't think you need this `if else`

components/ShareDog.js
```diff
@@ -0,0 +1,32 @@
...
+        }
+      } else if (result.action === Share.dismissedAction) {
+        
+      }
+    } catch (error) {
+      alert(error.message);
+    }
+  };
+  return (
+    <View style={{ marginTop: 0, marginBottom: 30 }}>
```
EmOlsz, line 28: Same as before, this should be styled component

EmOlsz (changes requested):

components/RandomDog.js
IngelaL: I don't know how to do that since this is an image component imported from react-native. I can't use the name Image as name in a styled component. Do you know how I can pass the styles as a styled component?

IngelaL:
@EmOlsz I have now made the changes requested. However, I don't know how to style the Image component as a styled component. See my comment above. /Ingela

components/RandomDog.js
EmOlsz: Yep, you can do it like this for example :

```
const DogImage = styled.Image`
    width: 270px;
    height: 270px;
`
```

EmOlsz:
Hey Ingela! Check out my comment :)

components/RandomDog.js
IngelaL: Ah of course, thanks! Didn't think I could change that Image name since it was imported, but the problem is fixed now :)

EmOlsz (approved):

PR #105, opened 2020-11-15
Breaking News

karanmann:
I though this weeks project was both Exciting and tough. 
Because one we were left in the wild and given the tools to survie and not knowing what to expect and two that we really had to comeup with our own projects.

I think react native is a fun language with a lot of area of coding to venture into and it is something that I would for sure explore. 

On the other hand it was frustrating initially with all the errors but as I dug in more and understood the documentation better it became much easier. 

Luckily I did not have much of error solving as I was approaching the problem one at a time.

I was able to make it multipage, Add Lottie Animation and navigate. So al in all I am pretty happy with the result.


App.js
```diff
@@ -1,26 +1,25 @@
...
 const App = () => {
   return (
-    <Container>
-      <Title>This is your cool app!</Title>
-      <Title>Go to App.js and start coding</Title>
-      <Title>💅💅💅</Title>
-    </Container>
+    <>
+    <NavigationContainer>
+      <Stack.Navigator initialRouteName="Home">
```
EmOlsz, line 14: Great job with Navigation :)

App.js
```diff
@@ -1,26 +1,25 @@
...
+const Stack = createStackNavigator();
 
 const App = () => {
   return (
-    <Container>
-      <Title>This is your cool app!</Title>
-      <Title>Go to App.js and start coding</Title>
-      <Title>💅💅💅</Title>
-    </Container>
+    <>
```
EmOlsz, line 12: No need to use React Fragment here since you return only one element (`<NavigationContainer />`)

components/Footer.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+import technigo from '../assets/technigo.png'
+import { FooterContainer,FooterText,TechnigoImage } from './Style/StyleFooter'
```
EmOlsz, line 4: folders shouldn't have capitalised names, `./style/StyleFooter` is more proper

components/LatestNews.js
```diff
@@ -0,0 +1,55 @@
...
+  ImageContainer, 
+  TitleText, 
+  AuthorName, 
+  TimedText, 
+  ReadMore,
+  ClockImage} 
+  from './Style/StyleLatestNews'
+
+import clock from '../assets/clock.png'
+import { ScrollView } from 'react-native-gesture-handler'
```
EmOlsz, line 18: This dependency should be moved above, to other outer dependencies

components/Style/StyleHome.js
```diff
@@ -0,0 +1,24 @@
...
+export const BannerImage = styled.Image`
+  align-items: center
+  resize-mode: contain
+  height: 50px
+  width: 50px
+`
+
+export const HeaderContainer = styled.View`
+align-items: center
+background: rgb(219,228,238)
```
EmOlsz, line 20: Wrong indentations

EmOlsz (approved):

PR #107, opened 2020-11-15
React-App -Claudia

ckjerrgren:
I would have liked to do something more advanced, but because I spent so much time with the technology this week it had to be something more simple. 

EXPO!!
- it really does not want to work for me.


app/components/Main.js
```diff
@@ -0,0 +1,131 @@
...
+  Text,
+  View,
+  TextInput,
+  ScrollView,
+  TouchableOpacity,
+} from 'react-native';
+
+import Note from './Note';
+
+export default class Main extends React.Component {
```
EmOlsz, line 13: Is there a reason why you used class-based component?

app/components/Note.js
```diff
@@ -0,0 +1,54 @@
...
+  Text,
+  View,
+  TouchableOpacity,
+} from 'react-native';
+
+export default class Note extends React.Component {
+  render() {
+    return (
+
+      <View key={this.props.keyval} style={styles.note}>
```
EmOlsz, line 13: No need for `key` attribute here, we do not perform any loop

EmOlsz (changes requested):

app/components/Main.js
ckjerrgren: I think endless hours looking at tutorials led me there.

ckjerrgren: I think endless hours watching tutorials led me there.

ckjerrgren:
the two comments should be fixed now

HIPPIEKICK (approved):

PR #111, opened 2020-11-15
Best seller book app

BamStugendal:
I actually think this week has gone well. Better than I thought. Of course, were there a lot of error messages that I didn't understand, but somehow it turned out well even with the struggle with that. 😄 

### DIFF
--- a/App.js
+++ b/App.js
@@ -21,22 +21,12 @@ const Body = styled.SafeAreaView`
 `
 
 
-
-function DetailsScreen() {
-  
-  return (
-    <Details />
-  );
-}
-
-
 const HomeStack = createStackNavigator();
 function HomeStackScreen() {
  return (
    <HomeStack.Navigator>
     <HomeStack.Screen name="Home" component={HomeScreen} />             
-    <HomeStack.Screen name="Details" component={Details} />
-    <HomeStack.Screen name="Book" component={Book} />         
+    <HomeStack.Screen name="Details" component={Details} />      
    </HomeStack.Navigator>
   );
 }
@@ -51,8 +41,6 @@ function SettingsStackScreen() {
     </SettingsStack.Navigator>
   );
 }
-
-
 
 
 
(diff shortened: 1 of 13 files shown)


App.js
```diff
@@ -1,26 +1,88 @@
...
+  flex: 1;
+  margin-top: 0;
 `
 
-const App = () => {
+//Home stack with fition and nonfiction books.
+const HomeStack = createStackNavigator();
+const HomeStackScreen = () => {
+ return (
+  <HomeStack.Navigator>
```
EmOlsz, line 25: Great job with Navigation :)

Nonetheless, all of those components should be moved to separate files

Book.js
```diff
@@ -0,0 +1,53 @@
...
+            activeOpacity={0.6}
+            underlayColor="#DDDDDD"
+            onPress={() => navigation.navigate('Details', { title, author, image, description, buyLink })}
+        >
+            <Container >
+                <Bookcover
+                    source={{ uri: `${image}` }}
+                    style={{
+                        width: 144,
+                        height: 227,
```
EmOlsz, line 43: Those styles should be moved to styled component

Details.js
```diff
@@ -0,0 +1,70 @@
...
+//Show book details and buy link.
+const Details = ({ route }) => {
+    const { title, author, image, description, buyLink } = route.params;
+    return (
+        <Container>
+            <Bookcover
+                    source={{ uri: `${image}` }}
+                    style={{
+                        width: 230,
+                        height: 400,
```
EmOlsz, line 56: Here as well

Details.js
```diff
@@ -0,0 +1,70 @@
...
+                    }}
+                />
+            <Title>{title}</Title>
+            <Author>{author}</Author>
+            <BookDescription>{description}</BookDescription>
+            <BuyButton 
+                title="Buy" 
+                onPress={()=>{ Linking.openURL(buyLink).catch((err) => console.error('An error occurred', err));}}
```
EmOlsz, line 64: It's pretty complicated logic, it should be moved to separate function

EmOlsz (changes requested):

EmOlsz (approved):

PR #113, opened 2020-11-15
Oscar Wilde quote-o-rama - Evelina 

evelinabe:
Had a lot of issues with the technique this week. Got a lot of warnings and errors about old versions of stuff that came with the project. When everything worked it was a charm, though. Had planned to implement an accelerometer and update quotes on shake but had to give that up and reinstall things instead.

I'd also like to have figured out how to set an initial quote and then hide it when pressing the touchable for the first time. Tried a few things but it didn't quite work, so hardcoded a quote instead.


App.js
```diff
@@ -1,26 +1,21 @@
-import React from 'react'
-import styled from 'styled-components/native'
+import React from "react";
+import HomeScreen from "./components/HomeScreen";
```
EmOlsz, line 2: Let's remember about one line break between outer and local dependencies

App.js
```diff
@@ -1,26 +1,21 @@
...
+import QuoteScreen from "./components/QuoteScreen";
 
-const Container = styled.View`
-  flex: 1;
-  background-color: papayawhip;
-  justify-content: center;
-  align-items: center;
-`
+import { NavigationContainer, StackActions } from "@react-navigation/native";
+import { createStackNavigator } from "@react-navigation/stack";
```
EmOlsz, line 6: And those should be moved above, to outer dependencies

App.js
```diff
@@ -1,26 +1,21 @@
...
 const App = () => {
   return (
-    <Container>
-      <Title>This is your cool app!</Title>
-      <Title>Go to App.js and start coding</Title>
-      <Title>💅💅💅</Title>
-    </Container>
-  )
-}
+    <NavigationContainer>
```
EmOlsz, line 12: Great job with Navigation :)

EmOlsz (changes requested):

EmOlsz (approved):

PR #117, opened 2020-11-16
Project React Native - Emma's Christmas app ☃️ 🎄

emma-tech:
I'm so happy with this project! Loved the creative freedom.  It was a real challenge though to make things work in React Native, and one week is maybe a little bit to short and stressful to really learn everything you need to know to build a project like this 😅 but it was fun, and  I'm very happy with the result 🤩

Could I've solved the navigation back to the startscreen in another way? I see some people are using navigation.goBack, is that a better approach maybe? :)


App.js
```diff
@@ -1,26 +1,29 @@
...
+import TipScreen from './components/TipScreen';
 
-const Title = styled.Text`
-  font-size: 24px;
-  color: palevioletred;
-`
+// Navigation
+import { NavigationContainer } from '@react-navigation/native';
+import { createStackNavigator } from '@react-navigation/stack';
+import 'react-native-gesture-handler';
```
EmOlsz, line 10: Thank you dividing dependencies by usage :) Nevertheless, those dependencies should be moved above local ones

App.js
```diff
@@ -1,26 +1,29 @@
...
 const App = () => {
   return (
-    <Container>
-      <Title>This is your cool app!</Title>
-      <Title>Go to App.js and start coding</Title>
-      <Title>💅💅💅</Title>
-    </Container>
-  )
-}
+    <NavigationContainer>
```
EmOlsz, line 16: Great job with Navigation!

EmOlsz (approved):

PR #119, opened 2020-11-16
Week 13 - Project React Native App - Johanna Rexin

johannarexin:

### DIFF
--- a/App.js
+++ b/App.js
@@ -1,26 +1,22 @@
 import React from 'react'
+import data from './data.json'
 import styled from 'styled-components/native'
-
-const Container = styled.View`
-  flex: 1;
-  background-color: papayawhip;
-  justify-content: center;
-  align-items: center;
-`
-
-const Title = styled.Text`
-  font-size: 24px;
-  color: palevioletred;
-`
+import { Training } from './components/Training'
 
 const App = () => {
+  
   return (
     <Container>
-      <Title>This is your cool app!</Title>
-      <Title>Go to App.js and start coding</Title>
-      <Title>💅💅💅</Title>
+      <Training activitiesArray={data.training}/>
     </Container>
   )
-}
+  }
+
+  const Container = styled.View`
+  flex: 1;
+  backgroundColor: #ffad57;
+  justifyContent: center;
+  alignItems: center;
+`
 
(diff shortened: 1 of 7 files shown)


App.js
```diff
@@ -1,26 +1,22 @@
 import React from 'react'
+import data from './data.json'
```
EmOlsz, line 2: Let's remember that `data` is local dependency, therefore it should be placed below outer dependencies

App.js
```diff
@@ -1,26 +1,22 @@
...
+      <Training activitiesArray={data.training}/>
     </Container>
   )
-}
+  }
+
+  const Container = styled.View`
+  flex: 1;
+  backgroundColor: #ffad57;
+  justifyContent: center;
```
EmOlsz, line 18: Wrong indentations

components/Training.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react'
+import { useState } from 'react'
```
EmOlsz, line 2: You could simplify this : `import React, { useState } from 'react';`

components/Training.js
```diff
@@ -0,0 +1,58 @@
...
+import { View, Vibration } from 'react-native'
+import styled from 'styled-components/native'
+
+const TouchButton = styled.TouchableOpacity`
+  alignSelf: center;
```
EmOlsz, line 7: We should use kebab-case names in styled components, not camelCase

components/Training.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  return (
+    <View>
+      <Title>
+        {activity}
+      </Title>
+      <TouchButton onPress={() => {
+      getActivity(); 
+      Vibration.vibrate();}}>
```
EmOlsz, line 53: If you need to pass more than one line of code to event handler, then whole logic should be moved to separate function

EmOlsz (changes requested):

johannarexin:
Hi Maks! Now I've done the changes and pushed a new version :) Merry Christmas!

EmOlsz:
Thank you Johanna, project approved :)

EmOlsz (approved):

PR #120, opened 2020-11-16
React Native App Moa Hammarbäck 

mhammarback:
this week was really hard. Still fun, but really hard. I do however really like creating an app but i would like to know React a little bit better i think.  

I would have appreciate if knowing that you do not need to use the simulator to start with the project earlier. maybe in the project brief even. I think a lot of people struggled with the set up even though we in the end did not need it.


App.js
```diff
@@ -1,26 +1,26 @@
...
+import React  from 'react'
 
-const Container = styled.View`
-  flex: 1;
-  background-color: papayawhip;
-  justify-content: center;
-  align-items: center;
-`
+import { NavigationContainer } from '@react-navigation/native';
+import { createMaterialTopTabNavigator } from '@react-navigation/material-top-tabs';
```
EmOlsz, line 5: Those are outer dependencies, no need to divide them from other outer dependencies

App.js
```diff
@@ -1,26 +1,26 @@
...
 const App = () => {
-  return (
-    <Container>
-      <Title>This is your cool app!</Title>
-      <Title>Go to App.js and start coding</Title>
-      <Title>💅💅💅</Title>
-    </Container>
+  
+return (
+  <NavigationContainer>
```
EmOlsz, line 15: Great job with Navigation :)

Components/HomeScreen.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+
+import { Container, Title, InfoText } from './HomeScreenStyling'
```
EmOlsz, line 3: This is local dependency, it should be moved below all outer dependencies

Components/StartScreen.js
```diff
@@ -0,0 +1,41 @@
...
+
+import { Container, Title, ResultText, ChallengeButton, ButtonText } from './StartScreenStyling'
+
+export const StartScreen = () => { 
+  const [challenge, setChallenge] = useState('')
+
+  const onClick = (event) => {
+    event.preventDefault()
+      setChallenge(challengeArray)
+      console.log(challenge)
```
EmOlsz, line 11: We should remove all console.logs before final push

EmOlsz (changes requested):

mhammarback:
Thank you for your feedback Maks! 
I have completed the changes you requested :)

EmOlsz (approved):

PR #123, opened 2020-11-19
Project 13: React Native app

Thessan:
**My reflections on how this week's project turned out:** I couldn't grasp how to make the fetch work properly which made this project take a very long time. It's difficult when I don't always understand the written answers given and there was no code coach sessions available until after the week had ended.

However, in the end I now have an app that works. I would had liked to get the problems out of the way sooner though so I also could have had a bit more time to work on styling.

[https://snack.expo.io/@thessan/fortune-froggy](https://snack.expo.io/@thessan/fortune-froggy)


Components/FortuneMessage.js
```diff
@@ -0,0 +1,39 @@
+import React, { useState, useEffect } from 'react'
+import styled from 'styled-components/native'
+
+const FortuneText = styled.Text`
+font-size: 30px;
+color: white;
```
EmOlsz, line 6: Wrong indentations

Components/FortuneMessage.js
```diff
@@ -0,0 +1,39 @@
...
+            return Math.floor(Math.random() * Math.floor(max))
+        }
+
+        const randomizeFortune = () => {
+            fetch(fortuneURL)
+                .then(response => response.json())
+                .then(json => {
+
+        const formattedArray = json.map(item => item.text)  
+        const randomInt = getRandomInt(formattedArray.length)
```
EmOlsz, line 25: Wrong indentations as well

EmOlsz (changes requested):

Components/FortuneMessage.js
Thessan: I've made a change in the code, however I'm not sure it's how you meant

EmOlsz: It's ok now ;)

EmOlsz (approved):

PR #125, opened 2020-11-20
React Native "Find Meal" App

destinyleia:
Not as I wanted my app to be, but close enough !

More about apis and native


components/MealList.js
```diff
@@ -0,0 +1,142 @@
+import React, { useState } from 'react'
+import { Vibration } from 'react-native'
+import styled from 'styled-components/native'
+
+import backgroundImage from '../assets/background-fruits.png'
+
+const MealsArray = [
```
EmOlsz, line 7: Capitalised names are reserved for components only, not arrays

components/MealList.js
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  return (
+    <Container>
+      <BackgroundImageContainer
+        source={backgroundImage}>
+        <ButtonContainer
+          onPress={() => {
+            setRandomMeals()
+            Vibration.vibrate()
```
EmOlsz, line 79: If we need to pass more than one line of code to event handler, then we should move whole logic to separate function

components/MealList.js
```diff
@@ -0,0 +1,142 @@
...
+  )
+}
+
+const BackgroundImageContainer = styled.ImageBackground`
+  flex: 1;
+  justify-content: center;
+`
+
+const MealContainer = styled.View`
+`
```
EmOlsz, line 100: I don't think you need this styled component

components/MealList.js
```diff
@@ -0,0 +1,142 @@
...
+
+
+const Container = styled.View`
+flex: 1;
+`
+
+const ButtonContainer = styled.TouchableOpacity`
+top:  100px;
+margin: 100px;
```
EmOlsz, line 118: Wrong indentations

EmOlsz (changes requested):

components/MealList.js
destinyleia: I think I changed it right now - please check line 69 + 70 and 78 in the component MealList :)

EmOlsz (approved):

PR #133, opened 2021-04-09
Shake sensor

Sartish:
This week was probably my hardest week so far 😅🤯 But also now the most satisfying to complete! Since you really see how powerful React Native really is! So I am really glad you included this in the bootcamp. And it feels also like something you can continue working on myself since there is really good documentation in expo!

 Maybe more about React native navigation, since this is a powerful tool in React native, didn't notice it exists at first. But on the other hand it is also good to look up these things myself. So maybe the short answer should be no haha!


components/AffirmationMessage.js
```diff
@@ -0,0 +1,72 @@
...
+  useEffect (()=> { 
+    fetch('http://www.boredapi.com/api/activity/')
+      .then (response => response.json())
+      .then ((json) => { 
+        setMessage(json)       
+      });
+  },[]);
+
+    return(
+      <>
```
KarolCislowski, line 59: you have only one Container here sou there is no need to use React fragment here.

App.js
```diff
@@ -1,25 +1,52 @@
...
+      <Stack.Navigator initialRouteName="Home">
+        <Stack.Screen name="Start" component={HomeScreen}
+        options={{
+          headerTitleStyle: {
+            color: 'thistle',
+          },
+          headerStyle: {
+            backgroundColor: '#833471',
+          },
+        }} />
```
KarolCislowski, line 24: You have 3 components, 9 lines each, and the only difference between them is 2 words. This is a great opportunity to create an array containing objects with these 2 fields for everyone option and build this logic with the map().

KarolCislowski (approved):
This is a great app. Very nicely styled and the code is clean and simple to understand 🦄 🔮

PR #136, opened 2021-04-09
Ylva Landoff Lindberg, React-Native

sssylvasss:
I missed that it was different branches with code already to start with. Put in the end it worked. I made the sensor. I think this could have been really fun, but I feel stressed because it was a lot of things I didn't understand and big problems setting up expo. 

But this week have still made me think that I really want to work more with mobile apps. 

I don't know how I missed that there was different branches but I wish that that had been pointed out very clear in the beginning of the week. That would have made everything a bit more easy:) 

I wish we had at least one more week with React-native


components/SensorImagePicker.js
```diff
@@ -0,0 +1,91 @@
...
+    require("../assets/aldis-popcorn-11.jpg"),
+  ]
+  const [popcorn, setPopcorn] = useState(null);
+  useEffect(() => {
+    !isShaking(data) && setPopcorn(popcornArray[Math.floor(Math.random()*popcornArray.length)]);
+    vibrateDevice();
+  }, [isShaking(data)]);
+
+  return (
+    <>
```
KarolCislowski, line 76: ShakeView is a single component here so there is no need to use React Fragment here.

KarolCislowski (approved):
Nice app and clean code. Remember to use <></> only when needed 🥇

PR #142, opened 2021-04-09
Shake sensor - Matilda Svensson

matildasvensson85:
It went ok, it was a little bit hard to understand react native at first.


KarolCislowski (approved):
Nice app 🥇

PR #144, opened 2021-04-09
Magic Ball_Irina Jekabsone

Irina-web-dev:

This weeks project was hard, mostly because we didn´t have clear instructions and explanations on how to, for example, use the sensors. I managed to implement navigation and detect a shake, but I don´t fully understand the code. I wish we could have a session explaining these things.


KarolCislowski (approved):
Great App 😸

PR #145, opened 2021-04-10
Shake sensor

katliebig:
My reflections on how this week's project turned out: I'm happy with it, even though I had to change from my initial idea.

Things I'd like to have clarified or explained in more detail: I know this project is built on the shake-sensor branch, but it's not using the shake sensor at all. 🤷‍♀️


KarolCislowski (approved):
Great app 🥇

PR #147, opened 2021-04-10
Project react native app - Andrea Larsson

Andrealarsson:
To be my first project in React Native I think it all went pretty smooth. I had some problem, in the beginning, viewing the app on my phone via Expo go. So I just looked on the browser which didn't look the same as on the phone and had to redo some styling.


components/Movie.js
```diff
@@ -0,0 +1,72 @@
...
+      `${title}`,
+      `Rating: ${rating} - ${overview}`,
+      [
+        { text: "Back to movies", onPress: () => console.log("Cancel Pressed")}
+      ]
+    );
+  }
+
+return (
+  <>
```
KarolCislowski, line 19: you are returning only TouchableOpacity here so why you are using extra fragments here??

components/MoviesList.js
```diff
@@ -0,0 +1,64 @@
...
+  const [movies, setMovies] = useState([])
+  
+  useEffect(() => {
+    fetch(URL)
+     .then(res => res.json())
+     .then(json => setMovies(json.results))
+  },[])
+
+  return (
+   <>
```
KarolCislowski, line 19: next misused react fragment here 😿

components/MoviesList.js
```diff
@@ -0,0 +1,64 @@
...
+    <Header>
+      <Title>MovieHunter</Title>
+      <MovieImage
+        source={movieicon} />
+    <Container>
+    <PopularText >Popular</PopularText >
+    {movies.map(movie => (
+      <Movie key={movie.id} id={movie.id} poster_path={movie.poster_path} title={movie.title} rating={movie.vote_average} overview={movie.overview}/>
```
KarolCislowski, line 27: if you are passing a lot of props then arrange them in the column to make it easier to read 🤓

KarolCislowski (approved):
Great app 🍿

PR #149, opened 2021-04-10
Shake sensor

julialeonieros:
I had some problems in the beginning getting expo to work, so a lot of time was spent on that and the app is therefore quite simple. If I had had more time I would have made an animation when the phone is shaken. Maybe also a third view that is shown as a start-screen so an answer isn't shown before the phone is shaken. 

The layout looks best on an iPhone 6/7/8


KarolCislowski (approved):
Nice app 🎱

PR #151, opened 2021-04-10
Project 13

AsaHildebrand:
This weeks project turned out quite well, but my time was very limited this week so I had to create a much more basic code than I wanted for this project.... 


components/CatSection.js
```diff
@@ -0,0 +1,68 @@
...
+        setSection(json)
+    })
+  }   
+     
+  useEffect(() => {
+    fetchSection()
+  }, [] )
+
+  return (
+  <>
```
KarolCislowski, line 55: There is no need to use react fragments if you are returning only 1 component.

KarolCislowski (approved):
Nice project 😸

PR #154, opened 2021-04-11
React Native App

Aepera:
This was a difficult project and I´m happy I managed to create it.

I set the size of the images in px, because with % they didn´t show up. But there must be a way to use %?


KarolCislowski (approved):
Nice project. Remember to use <></> only when needed 🌈

PR #155, opened 2021-04-11
React native app Johan Löfgren

johlofg:
A lot of trubleshooting this week with both expo go and Lottie but pretty satisfied with the final result.


KarolCislowski (approved):
nice app 🍹

PR #158, opened 2021-04-11
React native app Ulysses Veghelyi

Ulve17:

A pretty stressful week with very little time to spend making this weeks project butam satisfied with the project.


screens/ModelS.js
```diff
@@ -0,0 +1,59 @@
...
+          <Title>Model S</Title>
+          <SubTitle>Starting at $69,000</SubTitle>
+          <NavButton>
+            <Button title="Choose a different car" color='white' onPress={() => navigation.openDrawer()} />
+          </NavButton>          
+        </CarTitles>
+      </CarContainer>      
+    </FeedContainer>
+  );
+};
```
KarolCislowski, line 59: You created many components with the same structure. You should use a single component which would be filled with different data depending on chosen model.

KarolCislowski (changes requested):
Please make your code DRY by creating a reusable car component and passing data by props.

PR #159, opened 2021-04-11
Hannah Native App

jesinkey:
Im happy to having a "working" app with navigation and an API call. Im not done with styling however and will continue to complete this next week after the collab days. 

I thought this entire week was a little vague and would have like to have more guidance as we had to learn: native, expo, new style of debugging and new styling all in one week. Maybe split this over two weeks instead?


components/CardThumb.js
```diff
@@ -0,0 +1,43 @@
...
+  /*Deconstructing props*/
+  const { card, navigation } = props;
+
+  /*Function to create route for CardDetails when clicking the entire CardThumb*/
+  const onPressCardButton = () => {
+    navigation.navigate('CardDetails', { card: card });
+  };
+
+  return (
+    <>
```
KarolCislowski, line 30: you are returning a single component here, so there is no need to using react fragment.

KarolCislowski (changes requested):
Please add the link to the hosted app for testing.

components/CardThumb.js
jesinkey: Done!

jesinkey:
Hi! I have now added the published project in README

KarolCislowski (approved):
Great Project 😸

PR #161, opened 2021-04-11
React Native - Darcy

darcywinterhill:


components/MovieCard.js
```diff
@@ -0,0 +1,40 @@
...
+
+      ${vote_average} / 10`,
+      [
+        { text: "Back to movies"},
+      ]
+    );
+  }
+
+  return (
+    <>
```
KarolCislowski, line 29: You are returning a single component here so there is no need to use react fragment.

components/MovieList.js
```diff
@@ -0,0 +1,36 @@
...
+  
+  const [movies, setMovies] = useState([])
+  const API = 'https://api.themoviedb.org/3/movie/popular?api_key=e90c1322becf61268c0c0fea570c7099&language=en-US&page=1'
+
+  useEffect(() => {
+    fetch(API)
+    .then(res => res.json())
+    .then(json => {
+      setMovies(json.results)})
+    .catch(err => console.error(err))
```
KarolCislowski, line 24: Remember to remove all console outputs before requesting the pull.

KarolCislowski (changes requested):
Nice app but you need to clean your code a little bit.

KarolCislowski (approved):
🥳

PR #165, opened 2021-04-11
Siri Hemsley - React Native App

sirimatsdotter:
This week was really hard for me and I actually started over from scratch this weekend and build a completely different app than I started with. But In the end I'm actually pretty happy with what I achieved given the time I spent!

I believe I have been able to implement both navigation and API request with interaction. 

I'm super confused about how to apply the flex: 1/or other number on the children of a parent. I could not get it to work well. So at the moment my details page is not looking very good since my textwrapper had to have a set height because otherwise the children would not be covered by the black/transparent background for some reason.....


App.js
```diff
@@ -1,25 +1,66 @@
...
+        />
+        <Stack.Screen name="Upcoming" component={Upcoming} 
+        options={{
+          headerStyle: {
+            backgroundColor: 'black',
+          },
+          headerTintColor: '#fff',
+        }}
+        />
+        <Stack.Screen name="Detail" component={Detail}
```
KarolCislowski, line 52: I was not able to open the detail page of any of the movies 😭

KarolCislowski (changes requested):
Please fix the detail page 😸

App.js
sirimatsdotter: @KarolCislowski - I have now fixed the problem!

KarolCislowski (approved):
Nice project 😸

PR #167, opened 2021-04-11
Week 13 - Project react native app (shake sensor), Isabella Markskog

IsabellaM5:
**My reflections on how this week's project turned out**
I did the shake sensor application, which meet all the blue level requirements. 

This week's project was fun and challenging! React native is some pretty powerful stuff! I'm happy with how the project turned out! Well, at least up until when I was to publish it - I ran into some problems which I wasn't able to solve within the time frame of this project. So instead I created a snack on Expo for this project which you can view here: https://snack.expo.io/@isabellam5/project-react-native-app 

**Things I'd like to have clarified or explained in more detail**
Nothing at the moment when it comes to React Native in itself. 

What I would like to know is why I can't publish my project via Expo. I noticed there were a couple of warning messages, which I suspect is the cause for why I can't publish the project. But I can't quite make out where to even begin on debugging these warning messages...


KarolCislowski (approved):
Great app 🐲 🐺 🦌 🌹 🐙 🦁

PR #169, opened 2021-04-11
Shake sensor -React Native App By Linnéa Helén

Polkapetunian:
My reflections on how this week's project turned out: Ok

Things I'd like to have clarified or explained in more detail: I aimed for Blue goals, clearly.


KarolCislowski (changes requested):
You need to update the SDK used in your project to make it run on the actual version of the expo go.

PR #174, opened 2021-04-11
Martin D. - Zen Walker step-counter app

Martno96:
I'm happy with the app I've made, but sad that I couldn't get the build to publish correctly. Will fix as soon as I can figure it out.

No not really :)


KarolCislowski (changes requested):
You use an old version of SDK in your project. You need to update it because it is not possible to open it on the newest version of the expo go.

PR #176, opened 2021-04-11
Week 13, React Native App - Therese Hagelin

ThereseHag:
It has been a challenging week with lots of new things to consider and lot of installations required. Styled components was logic and fun to work with, but it took some time to grasp the general idea of working with React Native. Also had some struggles with the technical aspects with slow compiling, not being able to look at Android etc, so I combined working in expo dev tools and also creating a snack where I could look at the project in Android view. I've learnt a lot and had fun. 

I've noticed that the TouchableOpacity looks different in Android, iPhone and web, which is a bit confusing, so I'd like to understand more what causes this. The text in it doesn't show on iPhone after deploying (it looked fine before), but shows up just fine in Android and web. Would also like to find out more about how to share selected things fetched from API.

### DIFF
--- a/App.js
+++ b/App.js
@@ -1,11 +1,10 @@
 import React, { useState, useEffect } from 'react'
 import styled from 'styled-components/native'
-import axios from 'axios'
-import { View, Text, TouchableOpacity, ScrollView } from 'react-native'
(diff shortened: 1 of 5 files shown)


KarolCislowski (approved):
Nice app 🐩 🐶 🐕‍🦺

PR #195, opened 2021-12-03
Sabrina Zancotti compass React Native

Zancotti:
View it live:
https://expo.dev/@zancotti/zancotti-compass


components/Loading.js
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import styled from "styled-components/native";
+import LottieView from "lottie-react-native";
+
+// I only got Lottie to work when i followed Lotties instructions of how to use it in react native.
+// Hence there is a class function in this component, which means there wont be a useEffect.
```
isomoth, line 6: Good to know, thanks for the explanation! Interesting insight.

components/Loading.js
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import styled from "styled-components/native";
+import LottieView from "lottie-react-native";
```
isomoth, line 3: The loading Lottie is really pretty! It matches the design quite well.

components/Main.js
```diff
@@ -0,0 +1,123 @@
...
+    } else if (heading >= 112 && heading < 157) {
+      return "Southeast";
+    } else if (heading >= 157 && heading < 202) {
+      return "South";
+    } else if (heading >= 202 && heading < 247) {
+      return "Southwest";
+    } else if (heading >= 247 && heading < 292) {
+      return "West";
+    } else if (heading >= 292 && heading < 337) {
+      return "Northwest";
```
isomoth, line 47: Interesting to see how you implemented the cardinal points in a function! I was wondering if this could be expressed as a switch statement?

components/Main.js
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  // If there is no heading it shows an errorMessage.
+  if (!heading) {
+    if (errorMessage) {
+      return <Text> {errorMessage}</Text>;
+    }
+
+    // if the platform is a webbrowser a text that explains that it wont work on webbrowsers appear.
+    // Else it returns the loading container.
```
isomoth, line 58: That's a good pre-emptive approach! I hadn't even considered that use case.

components/Main.js
```diff
@@ -0,0 +1,123 @@
...
+  );
+};
+
+export default Main;
+
+// The padding top is adapted depending on the statusBarHeight on the mobile used.
+const Container = styled.View`
+  align-items: center;
+  justify-content: center;
+  padding-top: ${Constants.statusBarHeight}px;
```
isomoth, line 100: Good one! Another use case I wouldn't have considered.

isomoth (approved):
Awesome job Sabrina! I loved the sleek and clean styling. Less is more in this app! 

I was also great to see that you consider different use cases and manage error handling quite well. That's something I could learn a lot from :) 

I would've probably express the if-else statements from Main.js as a switch, but that's mostly syntactic sugar. 

Your app works just as it should, and the minimalist styling approach really suits the functionality. Good job!

components/Loading.js
Zancotti: 😍

components/Main.js
Zancotti: I thought about doing it with a switch but decided against it since i am doing the statement inbetween different values. I was unsure if switch could be used for this purpose in a easy way.

components/Main.js
Zancotti: Yeah it was very needed in my case😂 since it Didnt work on the browser at all.

Zancotti:
Thanks for a good review!

components/Main.js
isomoth: I see! Yeah, I haven't tried switch for that purpose either.

PR #204, opened 2021-12-03
Hedvigs trail app

HedvigM:
# Project React Native App 📱

In this project i wanted to try React Native out and se how i could use the sensors in the phone. I choose to build an App that gives you a new random trail suggestion when shaken.

## The problem

I thought it was so awesome to see the screen on the phone updated while I was writing my code on the computer. The first thing I did was print the x, y and z variable on the screen to see how the numbers changed when I touched the phone.

I had to be properly planned this week to get to the project due to the fact that I had so little time to spend this week. It was important that I put the time into the right things, so I started by planning what functions I wanted, then I ticked off one function after another.

If I had more time, I would make sure to get a nicer design. And above that I would like a back button, which sets the store to "null" so that you get back to the start page and can choose a new location.

## View it live

https://expo.dev/@hedvigm/trail-generator


Components/PackList.js
```diff
@@ -0,0 +1,61 @@
...
+import styled from 'styled-components/native';
+
+export const PackList = () => {
+  return (
+    <Container>
+      <Emojis> 🌲 🌲 🌲 </Emojis>
+      <InnerContainer>
+        <Text>
+          <Title>Packlist</Title>
+          {'\n'}
```
brunsant, line 14: Is this a way to make a line break? Haven't seen this before.

Components/Allemansrätt.js
```diff
@@ -0,0 +1,55 @@
...
+  text-align: center;
+  color: #1e5f18;
+`;
+
+const Title2 = styled.Text`
+  font-size: 15px;
+  text-align: left;
+  margin-bottom: 10px;
+
+  color: palevioletred;
```
brunsant, line 36: I realised that in almost all the styled components when added this particular color, there is a blank line on top of it.

brunsant (approved):
Amazing job this week Hedvig! 

I must say, this was one of the code reviews I learned most from!!! The design itself is really nice and I really enjoyed the color palette.
It's impressive to have the navigation and the sensor in the project, specially when you also added a redux store/reducer and a randomiser function. 

You should be really proud this week.

Components/PackList.js
HedvigM: Yes that's right, it's to make a break. I had to google it up, because the </br> don't work in react native.

Components/Allemansrätt.js
HedvigM: hmm that's weird, I don't know why I did that, but thank you for pointing it out =)

PR #208, opened 2021-12-04
Project React-native-app

ruruahn:

View it Live! : 
 exp://exp.host/@ruruahn/project-react-native-app


madeleinesvensson (commented):
Very nice written code! Very structured and easy to follow along thanks to naming and the comments, and also the indentations and the clustering of code(img together, accelerometer together etc)!! very nice! I couldn't open the app so I haven't seen it :( Maybe you can send me a link and I will check it out! from the looks of it's an office quote shaker. :P I really like the addition of the image depending on the person saying it but as commented maybe that code doesn't have to include all of the code that is repetitive throughout?  The quote component might be a bit long also(could be because of the conditional rendering ?) but maybe that could be broken into smaller components?  great job and I really want to see the app now!

PR #212, opened 2021-12-04
Linnéa Wilhelmsson week 16 

Skrosen:
I would have wanted to add more sensory functions but I'm still happy with the result. 

I'm not sure my loader works, i would have spent more time on this if i could this week.


App.js
```diff
@@ -1,26 +1,36 @@
-import React from 'react';
-import styled from 'styled-components/native';
+import React from "react";
+import { ImageBackground, StyleSheet } from "react-native";
```
annaester, line 2: I love all the thing s you can download to make react native (and react) work smoother. Like the ImageBackground

App.js
```diff
@@ -1,26 +1,36 @@
...
+			style={styles.background}
+		>
+			<GotApi />
+		</ImageBackground>
 	);
 };
 
 export default App;
+
+const styles = StyleSheet.create({
```
annaester, line 24: Nice that you used stylesheet and not styled components.

README.md
```diff
@@ -1,13 +1,16 @@
...
+Sharing or Multiscreen/Navigation.
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+I use onClick, subscribe, useEffect and accelerometer to make sure that the user
+can both press a button as well as shake the smartphone to update the quotes.
+With more time I would like to implement navigation and add more sensory
+features and also make sure that if the person that is quoted does not belong to
+any house the app renders only the name (without the comma after).
```
annaester, line 12: Aha, so the house was built in with the name? Annoying..

components/GotApi.js
```diff
@@ -0,0 +1,149 @@
...
+	StyleSheet,
+} from "react-native";
+import { Accelerometer } from "expo-sensors";
+
+const GotApi = () => {
+	const [quote, setQuote] = useState({});
+	const [loading, setLoading] = useState(false);
+
+	//communication between code and position of phone
+	const [subscription, setSubscription] = useState(null);
```
annaester, line 16: What exactly did the subscription do? Is it just connected to the shaking part? Simply "subscribing" the movement part of the app and rendering?

components/GotApi.js
```diff
@@ -0,0 +1,149 @@
...
+	const [data, setData] = useState({
+		x: 0,
+		y: 0,
+		z: 0,
+	});
+
+	//generates quote when button is clicked
+	useEffect(() => {
```
annaester, line 24: Why having it here? When you have the fetch below - you should just be able to call the function in your return (which you do on line 78) without having to use useEffects once again up here.. maybe this is why your loader doesn't work..?

annaester (commented):
Really great job! Very easy code to read and follow. I like that you used StyledSheets. And I do agree with the irritation of not being able to manage/change you respons from the API (like the name part).

PR #216, opened 2021-12-05
React native app Maria Petersson

hejmaria:
Fun introduction to React native. I both want and need more time to get really grasp the tool/language. 


components/ShakeApi.js
```diff
@@ -0,0 +1,126 @@
...
+  flex: 1;
+  justify-content: center;
+  align-items: center;
+`;
+
+const KanyeBackground = styled.ImageBackground`
+  height: 100%;
+`;
+
+const ShakeApi = () => {
```
Zancotti, line 46: I didnt do a shake app, so i think it is so interesting to see how it is built up.

components/ShakeApi.js
```diff
@@ -0,0 +1,126 @@
...
+    Poppins_400Regular,
+  });
+
+  useEffect(() => {
+    generateQuote();
+  }, []);
+
+  //   how often the acc will check + when component get mounted we will subscribe, and then unmount(?)
+  useEffect(() => {
+    Accelerometer.setUpdateInterval(1000);
```
Zancotti, line 68: So if im understanding this correctly it is updating every second to see if the cellphone is shaken?

Zancotti (commented):
Hey Maria, 

A very well executed project. The code was nice and well sorted. I have to admit react native wasnt my cup of tea and due to that I have a very hard time of actually giving som constructive feedback on your code since i actually dont know the best way to do it. I can just conclude that your code seems to work as intended and that you have done a great work!

Well done!

PR #233, opened 2021-12-05
Eddie-Lundgren-React-native-app

Lundgreneddie:
https://expo.dev/@zourak/project-react-native-app

### DIFF
--- a/components/Facts.js
+++ b/components/Facts.js
@@ -4,30 +4,62 @@ import {
   Text,
   TouchableOpacity,
   ActivityIndicator,
-  Pressable,
+  ImageBackground,
+  StyleSheet,
 } from "react-native";
 import styled from "styled-components/native";
 
+const image = {
+  uri: "https://images.unsplash.com/photo-1569982175971-d92b01cf8694?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=735&q=80",
+};
+
+const styles = StyleSheet.create({
+  background: {
+    flex: 1,
+    justifyContent: "center",
+  },
+});
+
+const Body = styled.View`
+  height: 100%;
+`;
+
 const FactContainer = styled.View`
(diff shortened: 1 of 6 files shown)


components/Facts.js
```diff
@@ -0,0 +1,99 @@
...
+export const Facts = () => {
+  const [fact, setFact] = useState({});
+  const [loading, setLoading] = useState(false);
+
+  useEffect(() => {
+    generateFacts();
+  }, []);
+
+  const generateFacts = () => {
+    setLoading(true);
```
katiewu1, line 74: Maybe can add a nice loading indicator when loading, or use the same background image? Right now the user get a white background between each time user click on the Generate-button. 
Have you tried using ternary operator before? So when state loading is true we want to do something, and the second operator we want to generate the fact. 
For example: 
{loading ? <loadingIndicator /> : (<FactContainer> ... </FactContainer>) }

When loading is true - go to loadingIndicator component
When loading is false - display the FactContainer

The ImageBackground can stay the same, and you can put the ternary operator inside of the ImageBackground. I think both for the loadingIndicator and the FactContainer will then have the same background image.

components/Facts.js
```diff
@@ -0,0 +1,99 @@
...
+    fetch("https://uselessfacts.jsph.pl/random.json?language=en")
+      .then(response => response.json())
+      .then(json => setFact(json))
+      .finally(() => setLoading(false));
+  };
+
+  if (loading) {
+    return <ActivityIndicator />;
```
katiewu1, line 82: Oh... I see that you've already an ActivityIndicator when loading is true 😅🤣
Nevermind then hehe... 
But I don't think I can review your ActivityIndicator, don't see that component here 🤔

katiewu1 (commented):
Hey Eddie, great job! I like your app! Now I know the longest word in English is pneumonoultramicro..... Not sure how to pronounce it 😆 And toilets flush in E flat... 🎶 
You have met the blue level requirements 👏 

I made some comments, hope you can find them useful. 
Enjoy the rest of the weekend!

PR #248, opened 2021-12-21
Shake sensor

Mattsson717:
It was a fun project, but I could'nt get the Expo go app to work.


components/Dogs.js
```diff
@@ -0,0 +1,128 @@
...
+	const [data, setData] = useState({
+		x: 0,
+		y: 0,
+		z: 0,
+	})
+	const [dog, setDog] = useState([])
+	const [quote, setQuote] = useState({})
+	const [loading, setLoading] = useState(false)
+	const [subscription, setSubscription] = useState(null)
+	const [fontsLoaded] = useFonts({
```
Ajliin, line 46: Nice made with google fonts. Is there a way to get that font everywhere? I notice now that u have to add it on each style.

Ajliin (commented):
Really nice job, I tried my best to find things to comment on but find very few. You have met every requirement and add a lot of extra with navigator bar and google font. A generell advise would be to try to comment a bit in the code, will make it easier to follow when it is this big components!

PR #261, opened 2022-04-28
Sensors QuoteShaker -Frida Axelsson

Nosslexa:
https://expo.dev/@nosslexa/project-react-native-app?serviceType=classic&distribution=expo-go

### DIFF
--- a/components/SensorComponent.js
+++ b/components/SensorComponent.js
@@ -2,95 +2,64 @@ import React, { useState, useEffect } from "react";
 import { Accelerometer } from "expo-sensors";
 import styled from "styled-components/native";
 
-// ==========================
-// = Functions
-const isShaking = (data) => {
-  // x,y,z CAN be negative, force is directional
-  // We take the absolute value and add them together
-  // This gives us the total combined force on the device
-  const totalForce = Math.abs(data.x) + Math.abs(data.y) + Math.abs(data.z);
+import { View, Text, TouchableOpacity } from "react-native";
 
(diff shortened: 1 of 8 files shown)


components/utils/urls.js
```diff
@@ -0,0 +1,2 @@
+
+export const URL = "https://api.quotable.io/random"
```
rawisou, line 2: I would personally declare this one url in the same file as where it is used in this case but I understand that in most cases, we would want to hide our api keys and by having it in a separate file, we can add the file in the .gitignore.

components/SensorComponent.js
```diff
@@ -2,95 +2,107 @@ import React, { useState, useEffect } from "react";
...
+  text-align: left;
+  width: 300px;
+  margin: 15px;
 `;
-const ShakeData = styled.Text``;
+const Author = styled.Text`
+  font-style: italic;
+  font-size: 15;
+`;
+const Button = styled.TouchableOpacity`
```
rawisou, line 33: Nice use of TouchableOpacity, I discovered it after I already added a bunch of buttons in my project :(

rawisou (commented):
Hi Frida,

Your app works as it should. The shake sensor works and I get a new code each time I either press the button or shake the phone.

Your variables have semantic names which make your code easy to read and follow.

I tested it on an iPhone 12 pro and the last paragraph seem to go a bit too low but that does not affect the functionality of your app at all! It's very tricky to style react native apps which a bunch of possible screen sizes.
IMG_0202

GREAT JOB!!!

PR #270, opened 2022-04-29
React native app

qwer-ctrl:


components/ShakeAnswer.js
```diff
@@ -0,0 +1,108 @@
...
+
+      const Refresh = () => {
+        setAnswer('')
+      }
+
+    return(
+        <View style={styles.container}>
+            <Text style={styles.text}>{answer}</Text>
+            <View style= {{height: 100}}></View>
+            <TouchableOpacity onPress={Refresh} accessibilityLabel="Restart from this button">
```
tiiliu, line 101: This is a nice touch that you have an accessibilityLabel for the button 👍

components/ShakeAnswer.js
```diff
@@ -0,0 +1,108 @@
...
+	  color: "white",
+	  fontSize: 20,
+	  fontWeight: "bold",
+	  textAlign: "center",
+	  backgroundColor: "#000000c0",
+    fontFamily: "monospace"
+	},
+  ButtonText: {
+    color: 'white',
+    fontFamily: "monospace"
```
tiiliu, line 21: When opening your project on mobile, I get a warning about Monospace not being a system font. I googled a little bit and found some answers saying that it's not supported as a default font in iOS. Just as info, don't think it's a problem because I guess it just goes with some other font instead 😊  I tried the fontninja to recognize the fonts, but it unfortunately doesn't work so well with the localhost server (I had to clone your repo to see the project and be able to do the review because of my iPhone 🙊 )

App.js
```diff
@@ -1,26 +1,68 @@
...
-		<Container>
-			<Title>This is your cool app!</Title>
-			<Title>Go to App.js and start coding</Title>
-			<Title>💅💅💅</Title>
-		</Container>
+		<View style={styles.container}>
+			<Text style={styles.title}>MAGIC 8-BALL</Text>
+			<View style= {{width: '100%', height: '50%',  alignItems: 'center'}}>
+				<ImageBackground source={require('./assets/8-ball.png')} resizeMode='contain' style={styles.image}>
+						<View style={styles.SmallView}></View>
```
tiiliu, line 56: I noticed that this SmallView is empty, you can add marginTop to SmallViewSecond with the same result but less code 😊

App.js
```diff
@@ -1,26 +1,68 @@
...
+						<View style={styles.SmallViewSecond}>
+						<Text style={styles.SmallText}>pssst...shake me!</Text>
+						<Text style={styles.text}>Here is your answer:</Text>
+						</View>
+				<ShakeAnswer />
+				</ImageBackground>
```
tiiliu, line 62: Fixing the indentation inside of this ImageBackground would make the code more readable 😊

App.js
```diff
@@ -1,26 +1,68 @@
...
 const App = () => {
 	return (
-		<Container>
-			<Title>This is your cool app!</Title>
-			<Title>Go to App.js and start coding</Title>
-			<Title>💅💅💅</Title>
-		</Container>
+		<View style={styles.container}>
+			<Text style={styles.title}>MAGIC 8-BALL</Text>
+			<View style= {{width: '100%', height: '50%',  alignItems: 'center'}}>
```
tiiliu, line 54: Perhaps this styling could be also in the stylesheet when you already have one made?

components/ShakeAnswer.js
```diff
@@ -0,0 +1,108 @@
...
+      })
+
+      const Refresh = () => {
+        setAnswer('')
+      }
+
+    return(
+        <View style={styles.container}>
+            <Text style={styles.text}>{answer}</Text>
+            <View style= {{height: 100}}></View>
```
tiiliu, line 100: I noticed you have an empty View here too with only height attribute. I tried it locally and adding marginBottom to the styles.text will do the same thing for you with less code 😊

tiiliu (commented):
Great job with the project! The 8-ball works and looks very cool! 🤗  It also fills the requirements. I left some additional comments throughout the code, hopefully you find them useful. Nice work! 😎

PR #271, opened 2022-04-29
W.16 React native app - Jennifer Wang Kurtto

jjwwkkmm:
Tried out React Native and got an overview over native components.

Could further try out more built-in components for future projects

Link:
https://expo.dev/@jenniferwvng/project-react-native-app?serviceType=classic&distribution=expo-go
(Notice, link does not work on iOS by default)

### DIFF
--- a/components/ButtonApi.js
+++ b/components/ButtonApi.js
@@ -1,10 +1,10 @@
 import React, { useEffect, useState } from 'react'
-import { View, Text, TouchableOpacity } from 'react-native'
+import { Text } from 'react-native'
 import styled from 'styled-components/native'
 
 const APIButton = styled.TouchableOpacity`
     background-color: rgb(255, 255, 255);        
-    width: 250%;        
+    width: 50%;        
     align-self: center;
     margin: 20px;
     padding: 10px;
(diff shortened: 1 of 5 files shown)


components/ButtonApi.js
```diff
@@ -0,0 +1,50 @@
+import React, { useEffect, useState } from 'react'
+import { Text } from 'react-native'
```
sofiaringstedt, line 2: Nice that you are not importing TouchableOpacity etc, I also noticed that you do not need to import them when you are declaring them in styled components!

components/ButtonApi.js
```diff
@@ -0,0 +1,50 @@
...
+const APIButton = styled.TouchableOpacity`
+    background-color: rgb(255, 255, 255);        
+    width: 50%;        
+    align-self: center;
+    margin: 20px;
+    padding: 10px;
+    border-radius: 25px;
+`
+
+const ParentContainer = styled.View`
```
sofiaringstedt, line 14: Like this naming, very clear!

sofiaringstedt (commented):
Nice looking and well functioning website, good job! The code itself is also clear and easy to follow, so I do not have any remarks but rather just a few comments and things I thought about! When looking at last weeks instructions, it is not very clear to me whether or not you need to add a share function in order to meet the blue requirements. So one suggestion from me would be to add this if you think it is needed. I found that expo share was easier to implement than expo sharing. I also thought that this guide was easy to follow: https://blog.logrocket.com/sharing-content-react-native-apps-using-react-native-share/

PR #272, opened 2022-04-29
ProgramerQuotes App

id4h4lling:
https://expo.dev/@h4lling/project-react-native-app?serviceType=classic&distribution=expo-go


App.js
```diff
@@ -1,26 +1,45 @@
...
-	align-items: center;
-`;
-
-const Title = styled.Text`
-	font-size: 24px;
-	color: palevioletred;
-`;
+import React from 'react'
+import { StyleSheet, View, Text, ScrollView } from 'react-native'
+import { StatusBar } from 'expo-status-bar'
```
Thereese, line 3: It seems like this isn´t used so it can be removed, just a detail:)

App.js
```diff
@@ -1,26 +1,45 @@
...
+    alignItems: 'center',
+    marginLeft: 30,
+    fontFamily: 'Optima-Bold'
+  },
+  quoteContainer: {
+    marginTop: 20,
+    alignItems: 'center',
+    padding: 40
+  },
+  scrollView: {
```
Thereese, line 40: How does the scrollView work here? I used it myself and had it on the container where the fetched text was so when it was very long I could scroll it. I´m not sure if I used it correctly though, so my comment here is more out of curiousity.

components/ApiFetch.js
```diff
@@ -0,0 +1,90 @@
+import React, { useEffect, useState } from 'react'
+import { StyleSheet, View, Text, TouchableOpacity, Share } from 'react-native'
+
+const ApiFetch = () => {
+  const [quote, setQuote] = useState({})
+
+  const onShare = async () => {
```
Thereese, line 7: Nice that you got the share-function to work!

package.json
```diff
@@ -8,16 +8,20 @@
     "eject": "expo eject"
   },
   "dependencies": {
+    "@expo-google-fonts/bodoni-moda": "^0.2.2",
```
Thereese, line 11: Is this where you imported the fonts? Really nice to make that work!!

Thereese (approved):
Despite the technical issues you managed to make an app that works really good! The share-function is fun and so is the programming-quotes theme😊  The styling is clean and the colors are soft, the code is also easy to follow. I think it´s a job WELL DONE!🤩

PR #278, opened 2022-04-29
Project React Native

Thereese:

I chose to make an app which fetches image and information about space and presents it with image and title. More info about the picture is shown in a textbox that is shown when clicking a button. Another button shows a new image with info. Added a share-function but didn´t manage to make it work. I think it´s because I have mapped the fetched information. If I had more time I would have tried another approach. I chose to style with CSS in Js.


components/facts.js
```diff
@@ -0,0 +1,136 @@
...
+      "https://api.nasa.gov/planetary/apod?api_key=luBC7hHhG19fUe5tSWnL0sTNWq7NhgCjFcFpqigI&count=1"
+    )
+      .then((response) => response.json())
+      .then((data) => setFact(data));
+  };
+
+  const onShare = async () => {
+    try {
+      const result = await Share.share({
+        message: `More about space`,
```
Katarina821, line 90: Since ur `const fact` from use State is an object i think you can access it like `` ${fact[0].title}`` an  ``${fact[0].explanation} `` if you want to share your data from the API. Don´t know if that it´s the best way of doing it but it seems to work when i tried it:)

components/facts.js
```diff
@@ -0,0 +1,136 @@
...
+            >
+              <Text style={styles.moreText}>Learn more..&nbsp;&#8595;</Text>
+            </TouchableOpacity>
+            <TouchableOpacity style={styles.button} onPress={onShare}>
+              <Text style={styles.moreText}>Share</Text>
+            </TouchableOpacity>
+          </View>
+          <View style={styles.container}>
+            <ScrollView style={styles.scroll}>
+              {visible && <Text style={styles.text}>{item.explanation}</Text>}
```
Katarina821, line 128: Nice solution with the scrollview!

Katarina821, line 128: Really liked the visible solution, that tought me something new!

Katarina821 (approved):
Congrats on your first app, good job! Really liked the learn more function, nicely done:) I took down ur code to test the sharefuction and have an idea on how to make it work, i left a comment about it. Great job this week:)!

PR #279, opened 2022-04-29
Sensors

Kras053:
I made an app that fetches random images from a fox-image-API either when you shake your phone or when you click on the button.
This was a challenging week so I wish I had had more time for more functions and styling. 

See project here: 
https://expo.dev/@louisalo/project-react-native-app?serviceType=classic&distribution=expo-go


components/ShakeApi.js
```diff
@@ -0,0 +1,105 @@
+import React, {useState, useEffect} from "react";
+import { View, Text, TouchableOpacity, Image, StyleSheet } from "react-native";
+import { Accelerometer } from "expo-sensors";
+
+const styles = StyleSheet.create({
```
amandatange, line 5: Nice job with the stylesheet styled components! I like the clean design which really lets the foxes shine!

components/ShakeApi.js
```diff
@@ -0,0 +1,105 @@
...
+            alignItems: "center",
+            padding: 10,
+            borderRadius: 25,
+        },
+        text: {
+            color: "white",
+        }
+});
+
+const url = `https://randomfox.ca/floof/`
```
amandatange, line 28: Love the choice of api, clever to use something so cute which already does half the job with getting the app to look good ;)

components/ShakeApi.js
```diff
@@ -0,0 +1,105 @@
...
+
+const isShaking = (data) => {
+    const totalForce = Math.abs(data.x) + Math.abs(data.y) + Math.abs(data.z);
+    return totalForce > 1.78;
+}
+
+
+useEffect(()=> {
+    if (isShaking(data)) {
+        generatePicture();
```
amandatange, line 86: if you want to add something more to the page maybe something like a random fox fact could be of interest? :) https://some-random-api.ml/animal/fox

components/ShakeApi.js
```diff
@@ -0,0 +1,105 @@
...
+        <View style={styles.container}>
+            <Image style={styles.tinyFrame} 
+                    source={{ uri: picture.image}}/>
+                
+
+                <TouchableOpacity
+                    style={styles.button}
+                    onPress={generatePicture}>
+                    <Text style={styles.text}>Or click me!</Text>
+                </TouchableOpacity>
```
amandatange, line 100: Smart to add a button as well for generating a new image! Makes it more accessible I would think? And also easier for testing since you can use it in webbrowser:)

amandatange (commented):
I think you've done a really great job with your very adorable fox generator app! There is a clear usecase for the app and it really delivers that in a nice way! I think you could incorporate more elements to it as well without it feeling like there too much happening but it also works great as it is currently! I'm sorry I couldn't really add anything constructive about the code I think you've done it in a great way and I also don't really understand the shake sensor thing hehe :) I think you definitely have achieved the requirements for this project! :)

PR #283, opened 2022-04-29
Project 16 - React Native 

EmmaaBerg:
My reflections on how this week's project turned out: It was fun to explore React Native and I'm happy with my result. Would like to continue working on this project later on. 


components/Start.js
```diff
@@ -0,0 +1,42 @@
...
+import LottieView from 'lottie-react-native'
+
+import animation from '../assets/animation.json'
+
+
+const Start = () => {
+
+    return (
+        <View style={styles.container}>
+            <LottieView source={animation} autoPlay />
```
jessand77, line 12: Nice Lottie animation :-)

components/Start.js
```diff
@@ -0,0 +1,42 @@
...
+            <Text style={styles.title}>Are you <Text style={styles.highlight}>bored</Text> ?</Text>
```
jessand77, line 13: I like the highlight style here.

components/Start.js
```diff
@@ -0,0 +1,42 @@
...
+        color: '#ED254E',
+        fontWeight: 'bold'
+    },
+    subtitle: {
+        fontSize: 18,
+        fontStyle: 'italic',
+        color: 'black',
+        marginTop: 70
+    },
+})
```
jessand77, line 40: The code doesn't look so advanced here but still the result looks elegant I think - very good job!

components/ActivityApi.js
```diff
@@ -0,0 +1,55 @@
+import React, { useState } from 'react'
+import { StyleSheet, View, Text, TouchableOpacity } from 'react-native'
+
+
+const ActivityApi = () => {
```
jessand77, line 5: Clean and easy-to-read code in this component as well :-)

jessand77 (commented):
Good job with your React Native project, Emma! Good file structure and code that is well written and easy to follow. As I already mentioned, the Lottie looks very nice and the idea of suggesting things to do when you are bored is also a good one that I like.

PR #294, opened 2022-05-01
Week 16 - React Native App by Jenny Fiskaare

JenFi72:

Lack of time and a lot of technical issues. Fun, but challenging to understand what to install. 


components/SensorComponent.js
```diff
@@ -0,0 +1,118 @@
...
+
+// = Styled components
+const StepContainer = styled.View`
+display:flex;
+flex-direction:column;
+margin-top:10px;
+`;
+
+const StepTitle = styled.Text`
+  font-size: 70px;
```
jiiinCho, line 14: Maybe this font-size is a little bit too large in mobile screen?

components/SensorComponent.js
```diff
@@ -0,0 +1,118 @@
...
+  color: #97BFB4;
+  text-align:left;
+  font-style:italic;
+  opacity:0.7;
+  margin-bottom:-30px;
```
jiiinCho, line 19: This can cause displaying the title text off from the mobile screen :)

components/SensorComponent.js
```diff
@@ -0,0 +1,118 @@
...
+
+export const SensorComponent= () => {
+  const [StepAvailability, setStepAvailability] =
+  useState("");
+
+  const [stepCounter, updateStepCounter] = useState(0)
+
+  useEffect(()=>{
+    subscribe()
+
```
jiiinCho, line 74: you can unsubscribe function by returning it, for example 
`return () => _unsubscribe();`

components/SensorComponent.js
```diff
@@ -0,0 +1,118 @@
...
+  },[])
+
+  const subscribe = () => {
+    const subscription=Pedometer.watchStepCount((result)=>{
+      updateStepCounter(result.steps);
+    })
+
+    const _unsubscribe = () => {
```
jiiinCho, line 82: to unsubscribe Pedometer.watchStepCount method, you can define const _unsubscribe function outside of const subscribe function! (now it is in const subscribe function block)

components/SensorComponent.js
```diff
@@ -0,0 +1,118 @@
...
+      subscription && subscription.remove();
```
jiiinCho, line 83: to be able to use subscription as state variable, you can use 'useState' hook and defined it in the begging of this SensorComponent scope :)  
`const [subscription, setSubscription] = useState(undefined)`

then you would need to set the method to update subscription state after line 79, for example
`setSubscription(subscription);`

components/SensorComponent.js
```diff
@@ -0,0 +1,118 @@
...
+      watchStepCount(null);
+    }
+    
+// Checks if pedometer function is available on device
+
+    Pedometer.isAvailableAsync().then(
```
jiiinCho, line 89: I briefly checked the document, https://docs.expo.dev/versions/latest/sdk/pedometer/, maybe this Pedometer.isAvailableAsync().then() method needs to be placed in useEffect method??!

package.json
```diff
@@ -8,14 +8,24 @@
     "eject": "expo eject"
   },
   "dependencies": {
+    "-": "^0.0.1",
+    "@expo-google-fonts/lato": "^0.2.2",
```
jiiinCho, line 12: Awesome you used custom font style!

package.json
```diff
@@ -8,14 +8,24 @@
...
     "minimist": "^1.2.6",
     "react": "17.0.1",
     "react-dom": "17.0.1",
-    "react-native": "0.64.2",
+    "react-native": "0.64.3",
+    "react-native-accurate-step-counter": "^2.0.4",
+    "react-native-shake": "^5.1.1",
+    "react-native-vertical-text": "^1.0.5",
     "react-native-web": "0.17.1",
     "styled-components": "^5.3.3"
```
jiiinCho, line 30: Super great that you tried styled-component in this project! :)

jiiinCho (commented):
Great project with a through attention in styling and using pedometer function built in mobile! I cloned your project and run it locally because my iPhone is not able to run the deployed link in Expo go app. Unfortunately my iPhone was not able to detect any steps (which I still don't know why it does not work) but probably you already tested step detecting function. The app runs without any errors, maybe because my iPhone has small screen that caused display some text off from the screen. Feel free to get back to me if any comments are unclear :) great work!

PR #295, opened 2022-05-01
Simple Step counter App in react native 

Sherin-Susan-Thomas:

Expo Snack link: https://snack.expo.dev/@sherin11/step-counter-app


App.js
```diff
@@ -1,26 +1,17 @@
-import React from 'react';
```
Kyparn, line 1: Hi Sherin, i have been looking over your app and the functions on and tried it out. 
But i had one problem, it did not count my step.

App.js
```diff
@@ -1,26 +1,17 @@
...
 
 const App = () => {
-	return (
-		<Container>
-			<Title>This is your cool app!</Title>
-			<Title>Go to App.js and start coding</Title>
-			<Title>💅💅💅</Title>
-		</Container>
-	);
+  return (
```
Kyparn, line 10: Nice and clean app.

components/PedometerScreen.js
```diff
@@ -0,0 +1,62 @@
...
+  useEffect(() => {
+    subscribe();
+    return () => _unsubscribe();
+  }, []);
+  const _unsubscribe = () => {
+    subscription && _subscription.remove();
+    subscription = null;
+  };
+  const subscribe = () => {
+    const subscription = Pedometer.watchStepCount((result) => {
```
Kyparn, line 39: On the useState pedometer on line 25 is not use anywhere when im looking on your code in VSC

Kyparn (commented):
Nice job with clean app, not so mutch cooler and u are keeping the code and clean as always :)

components/PedometerScreen.js
Sherin-Susan-Thomas: pedometer is for checking whether a pedometer is available on the device. It shows true on mobile and false in web browsers. I did not add it because it just shows a boolean in text

PR #296, opened 2022-05-01
Sofia Ringstedt React Native project

sofiaringstedt:
https://expo.dev/@sofiaringstedt/project-react-native-app?serviceType=classic&distribution=expo-go


Nosslexa (commented):
Great job with this weeks project! Your app works great on my phone and i like that you have been combining using both the route and the shaker!
Also well done with finding your own Api and to make the fetch work. 
Your code is easy to follow and i don't find any duplicates.

PR #297, opened 2022-05-01
Sensors for week 16

Kyparn:

This week was fun but hard, all ot of new info to grasp :) In the end i would have try to make a magic 8 ball if i had time.

My site ----> https://expo.dev/@kyparn83/simon-react-native-app


idanaslund (approved):
The design on your app is great! It's clean but fun, I like that. You have successfully made the quotes appear while shaking the phone but also with the push of a button, really good job! I see you styled it with styled components, did you try the Stylesheet also? I felt Stylesheet was a bit difficult. 

Your code follows the guidelines for good code and you have achieved more than blue level requirements by using both sensors and fetching information from an API! Maybe try stretching even further with a Lottie animation too? Great work this week!

PR #303, opened 2022-05-02
React native app

sk-phan:


kolkri (commented):
Really great job with the react native app. I really liked this Helsinki activities app and nice that you found this open API for that. The app styling is very nice and minimalistic which gives the content more attention and the sharing functionality was great and work at my phone at least!

PR #309, opened 2022-05-06
Techy API React Native

Laura-Sjolander:


App.js
```diff
@@ -1,24 +1,48 @@
...
+	padding: 10px;
 `;
 
 const App = () => {
 	return (
 		<Container>
-			<Title>This is your cool app!</Title>
-			<Title>Go to App.js and start coding</Title>
-			<Title>💅💅💅</Title>
+			<Technician source={require('./assets/technicians-at-work.jpg')} />
```
nadialefebvre, line 39: Great that you made it work!

App.js
```diff
@@ -1,24 +1,48 @@
 import React from 'react';
 import styled from 'styled-components/native';
+import TechyApi from './components/TechyApi';
+import { ImageBackground } from 'react-native';
```
nadialefebvre: You could remove this line, since it's not used.

components/TechyApi.js
```diff
@@ -0,0 +1,57 @@
+import React, { useState, useEffect } from 'react';
+import { View, Text, TouchableOpacity } from 'react-native';
```
nadialefebvre: `View` and `TouchableOpacity` are not used, so it's better to remove them here (you don't have to import them when they are styled with styled-components).

components/TechyApi.js
```diff
@@ -0,0 +1,57 @@
...
+`;
+
+const APIButton= styled.TouchableOpacity`
+    font-weight: 700;
+    width: 40%;
+    background-color: green;
+    margin-top: 20px;
+    border-radius: 10px;
+    border: 2px solid darkgreen;
+    padding; 5px;
```
nadialefebvre: Your padding isn't applied here because you have a semi-colon instead of a colon, should be `padding: 5px;`

components/TechyApi.js
```diff
@@ -0,0 +1,57 @@
...
+    color: white;
```
nadialefebvre: This property has no effect (you text on the button remains black), since you are trying to set a font color to a container. Try to apply this to `Text` inside of the container instead.

components/TechyApi.js
```diff
@@ -0,0 +1,57 @@
+import React, { useState, useEffect } from 'react';
+import { View, Text, TouchableOpacity } from 'react-native';
+import styled from 'styled-components/native';
+
+const TechyApi = () => {
+    const [techwords, setTechwords] = useState({})
+
+    const generateTechWords = () => {
```
nadialefebvre, line 8: Nice API, I should try to learn some sentences ;)

components/TechyApi.js
```diff
@@ -0,0 +1,57 @@
...
+
+const Container = styled.View`
+    flex: 1;
+	justify-content: center;
+	align-items: center;
+	margin: 20px;
+    width: 80%;
+    padding-left: 20px;
+    padding-right: 20px;  
+    `;
```
nadialefebvre: Maybe it's just github that does that, but it seems that the indentation isn't right in some places, like here.

nadialefebvre (commented):
Good job, Laura! Be proud of yourself, from no knowledge of react Native to an entire mobile app!

components/TechyApi.js
Laura-Sjolander: Its Github for sure, it doesn´t look like this in VSC :)

components/TechyApi.js
Laura-Sjolander: Thanks! It´s the little things...right 😊

App.js
Laura-Sjolander: 😊

